PR #110, opened 2026-04-20
DEV-554: Add catalog commands

ajaen-ai:


cmd/agents.go
```diff
@@ -626,6 +626,54 @@ Examples:
...
+
+		if len(args) == 0 {
+			agents, err := deployClient.ListCatalogAgents(cmd.Context())
+			if err != nil {
+				return err
+			}
+			return output.PrintAgentCatalog(out, agents)
+		}
+
+		agentId := strings.TrimSpace(args[0])
```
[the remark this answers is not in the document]
ajaen-ai: Solved

internal/clients/deployment_client.go
```diff
@@ -1685,3 +1685,88 @@ func (c *DeploymentClient) GetAgentLogs(
...
+type catalogAgentsResponse struct {
+	Agents []CatalogAgent `json:"agents"`
+}
+
+type CatalogAgent struct {
+	Id string `json:"id"`
+}
+
+type catalogVersionsResponse struct {
+	Versions []string `json:"versions"`
```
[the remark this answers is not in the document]
ajaen-ai: The server always orders the versions and returns most recent -> less recent

internal/output/agents.go
```diff
@@ -87,3 +87,33 @@ func PrintAgentDescribe(out io.Writer, agent *clients.DescribeAgentResponse) err
...
 	return nil
 }
+
+func PrintAgentCatalog(out io.Writer, agents []clients.CatalogAgent) error {
+	if len(agents) == 0 {
+		fmt.Fprintln(out, "No agents available.")
+		return nil
+	}
+
+	headers := []string{"AGENT ID"}
```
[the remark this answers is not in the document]
ajaen-ai: Yes, we will update it when the server returns more information.

cmd/agents.go
```diff
@@ -626,51 +626,76 @@ Examples:
...
+}
 
-		projectName, err := sess.ResolveProject(cfg.Project, agentProject)
+var agentSchemaCmd = &cobra.Command{
+	Use:   "schema",
+	Short: "Display the JSON Schema for agent configuration",
+	Long: `Fetch and display the current JSON Schema for the agent_config block.
+
+Examples:
+	  iai agents schema`,
```
sbullones-ai: all other examples are indented with 2 spaces but this one indents with a tab + 2 spaces

ajaen-ai: Solved

internal/clients/deployment_client.go
```diff
@@ -1685,3 +1685,88 @@ func (c *DeploymentClient) GetAgentLogs(
...
+		return nil, fmt.Errorf("catalog request failed with status %s", resp.Status)
+	}
+
+	var result catalogAgentsResponse
+	if err := json.NewDecoder(resp.Body).Decode(&result); err != nil {
+		return nil, fmt.Errorf("failed to decode catalog response: %w", err)
+	}
+
+	return result.Agents, nil
+}
```
sbullones-ai, line 1733: We are defining the CatalogAgent and catalogAgentsResponse structs this way:
```
type CatalogAgent struct {
    Id string `json:"id"`
}

type catalogAgentsResponse struct {
    Agents []CatalogAgent `json:"agents"`
}
```
The only reason for this is that the deployment-operator response wraps
the agent IDs in `{"id": "..."}`. We then pass CatalogAgent around to
other functions (`cmd/agents.go`, `internal/output/agents.go`) that only
need the agent IDs. With the current shape, if the server ever changes
the response format, we have to update every call site — not just the
decoder.

Note: ListCatalogAgentVersions already returns []string directly, so
this change restores symmetry between the two catalog methods.

We can collapse to a single unexported response type and decode directly
inside ListCatalogAgents:
```
type catalogAgentsResponse struct {
    Agents []struct {
        Id string `json:"id"`
    } `json:"agents"`
}

func (c *DeploymentClient) ListCatalogAgents(ctx context.Context) ([]string, error) {
    // ... request ...
    var result catalogAgentsResponse
    if err := json.NewDecoder(resp.Body).Decode(&result); err != nil {
        return nil, fmt.Errorf("failed to decode catalog response: %w", err)
    }
    ids := make([]string, len(result.Agents))
    for i, a := range result.Agents {
        ids[i] = a.Id
    }
    return ids, nil
}
```
That way we decouple the server response shape from the call sites.

sbullones-ai: If we make this change then PrintAgentCatalog shape goes from `PrintAgentCatalog(out io.Writer, agents []clients.CatalogAgent) error` to `PrintAgentCatalog(out io.Writer, agents []string) error`. Now PrintAgentCatalog and PrintAgentVersions have the exact same shape: empty-message check → build single-column rows → call PrintTable. Only the header and empty message differ.

We could extract a helper:
```
func printSingleColumnTable(out io.Writer, header, emptyMsg string, values []string) error {
    if len(values) == 0 {
        fmt.Fprintln(out, emptyMsg)
        return nil
    }

    rows := make([][]string, len(values))
    for i, v := range values {
        rows[i] = []string{v}
    }

    return PrintTable(out, []string{header}, rows)
}

func PrintAgentCatalog(out io.Writer, agents []string) error {
    return printSingleColumnTable(out, "AGENT ID", "No agents available.", agents)
}

func PrintAgentVersions(out io.Writer, agentId string, versions []string) error {
    return printSingleColumnTable(
        out,
        "VERSION",
        fmt.Sprintf("No versions found for agent %q.", agentId),
        versions,
    )
}
```

ajaen-ai: It is very probable that in the future we return more information than just the agents ids, for example a description of each agent. I think we should keep it as it is.

In the case of agent versions it could grow but it is not as clear.

I think this is ok for now.

sbullones-ai (approved):

PR #120, opened 2026-05-13
DEV-642: Add database commands

ajaen-ai:


ajaen-ai:
@claude review again

All validations are done on the server side, including mutually exclusive flags, ignore those

PR #124, opened 2026-05-14
Add iai skills CRUD subcommand

ppalacio-iai:
## Summary

Adds first-class CLI support for Copilot skills, mirroring the typed-prompt UX already provided for routines/policies/variables/glossaries/macros/agents. Skills are free-form markdown bundles plus a structured `config.skill` block (`{description, intents[]}`) consumed by `interactive-chat`'s `skill_loader` at agent-creation time to build the intent → skill table. Until now the backend already accepted the type and error messages even pointed at `iai skills create`, but the command itself didn't exist — closing that doc-trap.

## What changed

- **`cmd/skills.go`** (new) — registers `skills` via the existing `registerPromptType` factory with two skill-specific extensions: `--description` / `--intents` (populate `config.skill`) and `--body` (inline alternative to `--file`).
- **`cmd/prompt_types.go`** — extends `PromptTypeConfig` with optional `AllowInlineBody` and `BindCreateConfigFlags` hooks. Both default off, so existing typed types keep their exact pre-change flag set and request payload. New helper `readPromptContent` dispatches `--file` vs `--body`.
- **`internal/clients/api_client.go`** — adds `Config map[string]any` (with `omitempty`) to `CreatePromptBody` so the create/update payload can carry `config.skill`. Existing callers produce byte-identical JSON.
- **`internal/output/prompts.go`** — `PrintPromptDetail` now renders a `Config:` block when non-empty. Empty/absent config (every other typed prompt today) is unchanged.
- **Docs** — regenerated via `make docs`. Only `docs/iai.md` changed in-place (one index line); the eight `iai_skills*.md` files are net-new.
- **Version bump** — `0.27.0 → 0.28.0` (minor: new feature) in `.bumpversion.toml` + `cmd/root.go`.

## Compatibility

Verified that existing typed prompts are unaffected:

- All unit tests pass (`go test -count=1 ./...`).
- `iai {policies,routines,variables,glossaries,macros} create --help` flag set is byte-identical to before — no `--body`, no `--description`, no `--intents` leakage.
- `iai macros create --file ...` end-to-end against dev: works, output identical.
- `iai routines get <existing>` (sports-assistant): no spurious `Config:` block — `hasConfigPayload` returns false on empty config.
- `iai macros create` without `--file` still rejected with the same error message.
- Generic `prompts` and `agents` command trees are on different code paths and not touched.

## Test plan

- [x] `go test -count=1 ./...` passes
- [x] `go build ./...` clean
- [x] `make docs` runs cleanly and only adds skill-related doc files (+ one line in the index)
- [x] End-to-end against dev backend (`platform-development` org, `interactive-copilot` project):
  - [x] `iai skills create iai-cli-smoke-test --file ./skill.md --description "..." --intents "a,b"` → version 1, config.skill populated in response
  - [x] `iai skills list` shows it
  - [x] `iai skills describe iai-cli-smoke-test --label latest` round-trips description + intents
  - [x] `iai skills update iai-cli-smoke-test --body "..." --description "v2" --intents "x,y"` → version 2 via inline `--body`
  - [x] `iai skills versions iai-cli-smoke-test` shows `2, 1`
  - [x] `iai skills delete iai-cli-smoke-test -f` removes it
  - [x] `iai skills create my-test` rejected with "[file body] is required"
  - [x] `iai skills create my-test --file f --body b` rejected (mutually exclusive)
  - [x] `iai macros create m1` still rejected without `--file` (regression check)

## Out of scope

- Bundling the new CLI into the `interactive-chat` sandbox image — separate follow-up.
- No backend or runtime changes — the `skill` type, `skills_v0/` folder, and the skill_loader consumer are already in place.

### DIFF
--- a/cmd/prompt_types.go
+++ b/cmd/prompt_types.go
@@ -15,10 +15,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// ConfigFlagBuilder is returned by BindPromptConfigFlags to assemble the
-// "config" field of a create/update payload from per-command flag values at
-// run time. Returns nil when the caller supplied no config-related flags so
-// the field is omitted from the request body.
+// ConfigFlagBuilder assembles the payload's "config" field from flag values.
+// Returns nil if no config flags were set.
 type ConfigFlagBuilder func() map[string]any
 
 type PromptTypeConfig struct {
@@ -29,15 +27,8 @@ type PromptTypeConfig struct {
 	Long         string   // long description for the parent command
 	RouteSegment string   // API URL segment for type-specific routes, e.g. "routines"
 	HasSchema    bool     // whether this type supports the schema subcommand
-	// AllowInlineBody adds a --body flag as an alternative to --file on the
-	// create and update subcommands. When set, exactly one of --file or
-	// --body must be provided.
-	AllowInlineBody bool
-	// BindPromptConfigFlags optionally registers type-specific flags on the
-	// given create or update command (e.g. --description, --intents for
-	// skills) and returns a builder that materializes those flags into the
-	// payload's `config` field at run time. Called once per command so each
-	// create/update gets its own flag-bound variables.
+	// BindPromptConfigFlags registers type-specific flags on create/update
+	// and returns a builder for the payload's config field.
 	BindPromptConfigFlags func(cmd *cobra.Command) ConfigFlagBuilder
 	CreateLong            string // long description for the create subcommand
 	ListLong              string // long description for the list subcommand
@@ -108,7 +99,6 @@ This is a public endpoint and does not require authentication.`, ptCfg.Plural),
 func makeCreateCmd(ptCfg PromptTypeConfig) *cobra.Command {
 	var (
 		file    string
-		body    string
 		labels  []string
 		tags    []string
 		project string
@@ -131,9 +121,9 @@ func makeCreateCmd(ptCfg PromptTypeConfig) *cobra.Command {
 		out := cmd.OutOrStdout()
 		name := strings.TrimSpace(args[0])
 
-		content, err := readPromptContent(cmd, ptCfg, file, body)
+		content, err := os.ReadFile(file)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to read file %q: %w", file, err)
 		}
 
 		pCtx, apiClient, _, err := resolveProject(cmd.Context(), org, project)
@@ -143,7 +133,7 @@ func makeCreateCmd(ptCfg PromptTypeConfig) *cobra.Command {
 
 		payload := clients.CreatePromptBody{
 			Name:   name,
-			Prompt: content,
+			Prompt: string(content),
 			Labels: labels,
 			Tags:   tags,
 		}
@@ -169,14 +159,7 @@ func makeCreateCmd(ptCfg PromptTypeConfig) *cobra.Command {
 	}
 
 	cmd.Flags().StringVar(&file, "file", "", "Path to the file containing the prompt content")
-	if ptCfg.AllowInlineBody {
-		cmd.Flags().
-			StringVar(&body, "body", "", "Prompt content provided inline (alternative to --file)")
-		cmd.MarkFlagsOneRequired("file", "body")
-		cmd.MarkFlagsMutuallyExclusive("file", "body")
-	} else {
-		_ = cmd.MarkFlagRequired("file")
-	}
+	_ = cmd.MarkFlagRequired("file")
 	cmd.Flags().
 		StringSliceVar(&labels, "labels", nil, "Labels for the prompt version (comma-separated)")
 	cmd.Flags().StringSliceVar(&tags, "tags", nil, "Tags for the prompt (comma-separated)")
@@ -184,29 +167,6 @@ func makeCreateCmd(ptCfg PromptTypeConfig) *cobra.Command {
 	cmd.Flags().StringVarP(&org, "organization", "o", "", "Organization name that owns the project")
 
 	return cmd
-}
-
-// readPromptContent returns the prompt body, resolving --file vs --body per
-// the type's AllowInlineBody setting. Cobra's MarkFlagsOneRequired /
-// MarkFlagsMutuallyExclusive already validate that exactly one of the two
-// is set when inline bodies are allowed, so this only needs to dispatch and
-// validate that the chosen flag is non-empty.
-func readPromptContent(
-	cmd *cobra.Command,
-	ptCfg PromptTypeConfig,
-	file, inline string,
-) (string, error) {
-	if ptCfg.AllowInlineBody && cmd.Flags().Changed("body") {
-		if strings.TrimSpace(inline) == "" {
-			return "", fmt.Errorf("--body must not be empty")
-		}
-		return inline, nil
-	}
-	content, err := os.ReadFile(file)
-	if err != nil {
-		return "", fmt.Errorf("failed to read file %q: %w", file, err)
-	}
-	return string(content), nil
 }
 
 func makeListCmd(ptCfg PromptTypeConfig) *cobra.Command {
@@ -320,7 +280,6 @@ func makeGetCmd(ptCfg PromptTypeConfig) *cobra.Command {
 func makeUpdateCmd(ptCfg PromptTypeConfig) *cobra.Command {
 	var (
 		file    string
-		body    string
 		labels  []string
 		tags    []string
 		project string
@@ -343,9 +302,9 @@ func makeUpdateCmd(ptCfg PromptTypeConfig) *cobra.Command {
 		out := cmd.OutOrStdout()
 		name := strings.TrimSpace(args[0])
 
-		content, err := readPromptContent(cmd, ptCfg, file, body)
+		content, err := os.ReadFile(file)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to read file %q: %w", file, err)
 		}
 
 		pCtx, apiClient, _, err := resolveProject(cmd.Context(), org, project)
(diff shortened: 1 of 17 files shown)


cmd/prompt_types.go
```diff
@@ -103,56 +120,87 @@ func makeCreateCmd(ptCfg PromptTypeConfig) *cobra.Command {
...
-	_ = cmd.MarkFlagRequired("file")
 
 	return cmd
 }
 
+// readPromptContent returns the prompt body, resolving --file vs --body per
+// the type's AllowInlineBody setting. Cobra's MarkFlagsOneRequired /
+// MarkFlagsMutuallyExclusive already validate that exactly one is set when
+// inline bodies are allowed, so this only needs to dispatch.
+func readPromptContent(ptCfg PromptTypeConfig, file, inline string) (string, error) {
```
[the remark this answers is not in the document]
ppalacio-iai: Fixed: `readPromptContent` now uses `cmd.Flags().Changed("body")` to detect explicit flag provision, then rejects with `"--body must not be empty"` if the value is blank or whitespace-only.

cmd/skills.go
```diff
@@ -0,0 +1,121 @@
...
+
+Examples:
+  iai skills describe summarize-trace
+  iai skills describe summarize-trace --version 3
+  iai skills describe summarize-trace --label staging`,
+		UpdateLong: `Update a skill by creating a new version with updated content.
+
+This creates a new version of the skill using the content from the provided
+file or --body string. --description and --intents are written into the new
+version's config.skill block; omit them to leave the previous version's
```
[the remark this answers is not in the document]
ppalacio-iai: Fixed: `UpdateLong` now explicitly states that each update creates a brand-new version with exactly the content and config provided — the previous version is not inherited. Users are told to re-pass `--description`/`--intents` on every update.

cmd/skills.go
```diff
@@ -0,0 +1,121 @@
...
+	var (
+		description string
+		intents     []string
+	)
+	cmd.Flags().StringVar(
+		&description, "description", "",
+		"Short description of the skill (stored in config.skill.description)",
+	)
+	cmd.Flags().StringSliceVar(
+		&intents, "intents", nil,
```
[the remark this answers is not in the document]
ppalacio-iai: Fixed: switched to `StringArrayVar` (repeated-flag only, no comma tokenisation). Updated all examples to use repeated `--intents` flags and added a unit test case that asserts a comma inside a single intent value is preserved.

cmd/prompt_types.go
```diff
@@ -23,11 +29,21 @@ type PromptTypeConfig struct {
...
+	// AllowInlineBody adds a --body flag as an alternative to --file on the
+	// create and update subcommands. When set, exactly one of --file or
+	// --body must be provided.
+	AllowInlineBody bool
+	// BindCreateConfigFlags optionally registers type-specific flags on the
+	// given create or update command (e.g. --description, --intents for
+	// skills) and returns a builder that materializes those flags into the
+	// payload's `config` field at run time. Called once per command so each
+	// create/update gets its own flag-bound variables.
+	BindCreateConfigFlags func(cmd *cobra.Command) ConfigFlagBuilder
```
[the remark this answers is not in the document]
ppalacio-iai: Fixed: renamed to `BindPromptConfigFlags` everywhere (field declaration, both call sites, and the `ConfigFlagBuilder` type comment).

cmd/skills.go
```diff
@@ -0,0 +1,147 @@
...
+		Long: `Manage Copilot skills for the interactive-chat service.
+
+IMPORTANT: These are Copilot skills — they are NOT interactive-agent behaviors.
+Skills here are loaded by the Copilot runtime (interactive-chat) and injected
+into Copilot conversations as context. They have no effect on the conversational
+agent (interactive-agent). To configure the agent use:
+  iai routines    — conversation flow scripts
+  iai policies    — behavioral guardrails
+  iai glossaries  — domain terminology
+  iai macros      — reusable text snippets
```
ajaen-ai: I think it is clear and we can avoid the "To configure..." and referencing all objects that configure the interactive agent.

cmd/prompt_types.go
```diff
@@ -23,11 +29,21 @@ type PromptTypeConfig struct {
 	Long         string   // long description for the parent command
 	RouteSegment string   // API URL segment for type-specific routes, e.g. "routines"
 	HasSchema    bool     // whether this type supports the schema subcommand
-	CreateLong   string   // long description for the create subcommand
-	ListLong     string   // long description for the list subcommand
-	GetLong      string   // long description for the describe subcommand
-	UpdateLong   string   // long description for the update subcommand
-	DeleteLong   string   // long description for the delete subcommand
+	// AllowInlineBody adds a --body flag as an alternative to --file on the
```
ajaen-ai: Suggest dropping `body` and `AllowInlineBody` entirely. Real skill content is multi-line markdown that's awkward to pass as a shell string. The PR's own examples had to be removed for exactly this reason. `file` already covers scripting; if "no temp file" is a real need, stdin support (`iai skills create foo --file - <<< "content"`) is the conventional Unix answer and needs no new flags or factory plumbing.

cmd/skills.go
```diff
@@ -0,0 +1,147 @@
...
+	registerPromptType(PromptTypeConfig{
+		TypeName:        "skill",
+		Plural:          "skills",
+		Aliases:         []string{"skill"},
+		Short:           "Manage Copilot (interactive-chat) skills — NOT interactive-agent behaviors",
+		AllowInlineBody: true,
+		Long: `Manage Copilot skills for the interactive-chat service.
+
+IMPORTANT: These are Copilot skills — they are NOT interactive-agent behaviors.
+Skills here are loaded by the Copilot runtime (interactive-chat) and injected
```
ajaen-ai: Is "interactive-chat" the final name? I thought we have settled on interactive-copilot.

On another note, I think the content regarding these are copilot skills and not interactive-agent behaviours is necessary only in the parent skills command. I don't think we need it in all the child commands (create, update...)

cmd/prompt_types.go
```diff
@@ -15,6 +15,12 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// ConfigFlagBuilder is returned by BindPromptConfigFlags to assemble the
+// "config" field of a create/update payload from per-command flag values at
+// run time. Returns nil when the caller supplied no config-related flags so
+// the field is omitted from the request body.
```
ajaen-ai: Suggest shortening: `// ConfigFlagBuilder assembles the payload's "config" field from flag values. Returns nil if no config flags were set.`

cmd/prompt_types.go
```diff
@@ -23,11 +29,21 @@ type PromptTypeConfig struct {
...
-	GetLong      string   // long description for the describe subcommand
-	UpdateLong   string   // long description for the update subcommand
-	DeleteLong   string   // long description for the delete subcommand
+	// AllowInlineBody adds a --body flag as an alternative to --file on the
+	// create and update subcommands. When set, exactly one of --file or
+	// --body must be provided.
+	AllowInlineBody bool
+	// BindPromptConfigFlags optionally registers type-specific flags on the
+	// given create or update command (e.g. --description, --intents for
+	// skills) and returns a builder that materializes those flags into the
```
ajaen-ai: Suggest shortening: `// BindPromptConfigFlags registers type-specific flags on create/update and returns a builder for the payload's config field.`

cmd/skills.go
```diff
@@ -0,0 +1,147 @@
...
+  iai skills delete summarize-trace --version 3
+  iai skills delete summarize-trace --label staging`,
+	})
+}
+
+// bindSkillConfigFlags registers --description and --intents for Copilot skill
+// create/update commands and returns a builder that assembles them into the
+// payload's config.skill block. These flags configure the Copilot runtime
+// (interactive-chat) only — they have no relation to interactive-agent
+// behaviors. Wire shape: config = {"skill": {"description": "...", "intents": [...]}}.
```
ajaen-ai: Suggest shortening: `// bindSkillConfigFlags registers --description and --intents and builds the config.skill payload block.`

internal/output/prompts.go
```diff
@@ -67,6 +67,15 @@ func PrintPromptDetail(out io.Writer, prompt *clients.PromptDetail) error {
 		fmt.Fprintf(w, "Updated At:\t%s\n", LocalTime(prompt.UpdatedAt))
 	}
 
+	if hasConfigPayload(prompt.Config) {
+		var indented bytes.Buffer
+		if err := json.Indent(&indented, prompt.Config, "", "  "); err == nil {
+			fmt.Fprintln(w)
```
ajaen-ai: The JSON block is inconsistent with the rest of the CLI output, which uses tabwriter key-value pairs (agents, services, databases, routines all do this). Suggest rendering config fields as flat key-value rows instead:

```
Description:   test skill
Intents:       greet, say hi
```

This keeps the metadata block uniform and the only visual break before `Content:`.

cmd/skills.go
```diff
@@ -0,0 +1,127 @@
...
+
+import "github.com/spf13/cobra"
+
+func init() {
+	registerPromptType(PromptTypeConfig{
+		TypeName: "skill",
+		Plural:   "skills",
+		Aliases:  []string{"skill"},
+		GroupID:  groupCopilot,
+		Short:    "Manage Copilot (interactive-copilot) skills — NOT interactive-agent behaviors",
```
ajaen-ai: Wdyt about:

`Manage Interactive Copilot skills (not to be confused with context items that configure the Interactive Agent)`

cmd/skills.go
```diff
@@ -0,0 +1,127 @@
...
+		Long: `Manage Copilot skills for the interactive-copilot service.
+
+IMPORTANT: These are Copilot skills — they are NOT interactive-agent behaviors.
```
ajaen-ai: Proposal:

```
IMPORTANT: These are Interactive Copilot skills, not to be confused with
  context items that configure the Interactive Agent. Skills are loaded by the
  Copilot runtime and injected into Copilot conversations as context. They
  have no effect on the Interactive Agent.
```

ajaen-ai (approved):

PR #127, opened 2026-05-19
feat(prompts): add `labels set` command for prompt versions

amadrazo-iai:
## Summary

  Adds `iai <type> labels set <name> --version N --labels ...` for promoting an existing prompt version to a label without  creating a duplicate version. Wires into existing backend `PATCH .../versions/{version}` endpoints — **no backend changes**.

  Supersedes #97 (open since 2026-04-01, marked conflicting after the `cmd/prompts.go` / `cmd/prompt_types.go` rename + the addition of `describe`/`diff`/`versions` subcommands). Same feature, clean re-implementation against current `main`.

  ## Coverage

  All 7 prompt surfaces:
  - Generic: `iai prompts labels set <name>`
  - Typed (via `registerPromptType()`): `routines`, `policies`, `glossaries`, `variables`, `macros`, `skills`

  Note: `agents` is no longer wired through `registerPromptType()` (moved to the deployment-operator path in #104), so it's not in scope.

  ## Usage

  ```bash
  iai prompts labels set greeting --version 3 --labels production
  iai routines labels set my-routine --version 1 --labels staging,canary
  iai skills labels set my-skill --version 2 --labels production
  ```

  Labels are unique per prompt — assigning a label to one version atomically removes it from any other version that had it (server-side guarantee).

  ## Changes

  - **`internal/clients/api_client.go`** — new `SetLabelsBody` struct + `SetPromptLabels` method following the same 
  `newRequest` + manual marshal + envelope-unwrap pattern as `CreatePrompt`. 4 focused tests in `api_client_test.go` 
  covering generic + typed paths, body shape, and server-message surfacing on non-2xx.
  - **`cmd/prompt_types.go`** — adds `makeLabelsCmd(ptCfg)` (bare parent group with `label` alias) and
  `makeLabelsSetCmd(ptCfg)`, wired into `registerPromptType()` alongside `versionsCmd` / `diffCmd`. All 6 typed commands
  inherit automatically.
  - **`cmd/prompts.go`** — `makeGenericLabelsCmd()` / `makeGenericLabelsSetCmd()` for the generic prompts command.
  - **`docs/`** — 14 new auto-generated doc files (`*_labels.md`, `*_labels_set.md` for all 7 surfaces) + 7 updated parent
  indexes.

  The `labels` parent is intentionally a noun group (rather than a flat `set-labels` verb) so future label operations (`add`
   / `remove` / `list`) can land under it without breaking changes.

  ## Behaviour

  - Pure passthrough: no client-side validation of `--version` or `--labels` — backend decides what's valid, consistent with every other prompt command.
  - `--version` and `--labels` are both required; missing them exits non-zero with cobra's standard required-flag error.
  - Output uses the same `output.PrintPromptDetail` renderer as `create` / `get` / `update`.

### DIFF
--- a/cmd/prompt_types.go
+++ b/cmd/prompt_types.go
@@ -60,8 +60,18 @@ func registerPromptType(ptCfg PromptTypeConfig) {
 
 	versionsCmd := makeVersionsCmd(ptCfg)
 	diffCmd := makeDiffCmd(ptCfg)
+	labelsCmd := makeLabelsCmd(ptCfg)
 
-	parentCmd.AddCommand(createCmd, listCmd, getCmd, updateCmd, deleteCmd, versionsCmd, diffCmd)
+	parentCmd.AddCommand(
+		createCmd,
+		listCmd,
+		getCmd,
+		updateCmd,
+		deleteCmd,
+		versionsCmd,
+		diffCmd,
+		labelsCmd,
+	)
 
 	if ptCfg.HasSchema {
 		schemaCmd := makeSchemaCmd(ptCfg)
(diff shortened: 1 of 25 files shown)


amadrazo-iai:
@claude 

  Addressed one, deferring the other two with rationale.

  **1. Path construction style** — fixed in d161c12. Collapsed the formatter-mangled `+` chain into `fmt.Sprintf` with
  vertical args. Reads cleanly and survives `golines`.

  **2. `--version 0` validation** — declining. This is a deliberate design choice carried over from the PRD (FR-10: *"The 
  CLI does **not** validate `version > 0` or filter empty label strings client-side — all input passes through to the 
  backend."*) and from the original PR #97's history: I added this exact guard in
  [`d62b2af`](https://github.com/Interactive-AI-Labs/interactive-cli/commit/d62b2af), then explicitly reverted it in
  [`45bcb6d`](https://github.com/Interactive-AI-Labs/interactive-cli/commit/45bcb6d) with the rationale *"keep CLI as thin backend layer."*

  Codebase consistency backs that up: `makeGetCmd` and `makeDeleteCmd` both accept `--version` as a plain `IntVar` and pass it through without validation. Adding a guard only on `labels set` would make this command an outlier. The backend's error   (`Prompt not found: '<name>' version 0`) is clear enough — verified during the live smoke test against dev, where `--version 999` returned that exact message.

  If we want clearer "you probably meant a positive version" UX, the right place is a shared validator across all three
  `--version`-accepting commands, not just `labels set` — happy to do that in a follow-up if there's appetite.

  **3. `Args: cobra.NoArgs` on `labels` group** — declining. None of the existing prompt parent groups (`prompts`,
  `routines`, `policies`, `glossaries`, `variables`, `macros`, `skills`) use `cobra.NoArgs`. Adding it just to `labels`
  would create the same kind of one-off inconsistency that #2 would. Cobra's default behaviour for a group with subcommands is fine here ("unknown command" on extras).

  Re-review.

PR #133, opened 2026-05-26
Adjust output timestamp and secret key formatting

ajaen-ai:
## Summary
- switch LocalTime output to a compact local timestamp format
- reduce visible secret keys in secret list output from 3 to 2
- update output tests to match the new formatting

## Testing
- go test ./...


ajaen-ai:
Addressed Claude Issue 1 in 5bcdde5 by renaming the outdated test case. Leaving Issue 2 for now since `PrintSecretList` already exercises the production path that uses `maxVisible=2`.

PR #135, opened 2026-05-29
feat(integrations): add iai integrations MCP connection commands

ppalacio-iai:
## Overview

Adds a new `iai integrations` command group (aliased `mcp` / `integration`, under the **Infrastructure** group) for managing Model Context Protocol (MCP) connections in a project. Three-layer split matching the existing eval/agents/secrets structure:

- `cmd/integrations.go` — cobra commands
- `internal/clients/mcp.go` — thin HTTP client (reuses `doGet`/`doCreate`/`doDelete`, `evalBasePath`)
- `internal/output/integrations.go` — output printers (reuse `PrintTable`/`NewDescribeWriter`/`LocalTime`/`TruncateList`)

## Commands

| Command | Purpose |
|---|---|
| `list` / `get` | List connections; show one connection + its cached tools |
| `catalog` | Browse the curated MCP server catalog |
| `create-custom` | Connect a user-defined MCP endpoint |
| `create-from-catalog` | Connect a server from a catalog entry |
| `verify` | Re-dial the server and refresh the cached tool list |
| `delete` | Remove a connection (with y/N confirmation, `-f` to skip) |
| `tools run` | Invoke a tool with JSON args (`--args` / `--args-file`) |

## Robustness & security

- **EOF-safe delete confirmation** — honors input without a trailing newline; empty stdin safely declines instead of erroring.
- **`--credential-stdin`** on both create commands so secrets stay out of the process list / shell history (mutually exclusive with `--credential`).
- **Path-injection safe** — all dynamic path segments go through `url.PathEscape`.
- **Write-only credentials** — never returned by the API or printed (only `HasCredential`).
- `tools run` rejects non-object JSON args (including `null`).

## Testing

Unit coverage across all three layers: client methods against `httptest` (method/path/body asserted, including the catalog body and nil-args → `{}`), validation helpers, JSON/file arg parsing, credential resolution, and the output printers. `go build`, `go vet`, `gofmt`, and the full suite pass.

## Notes

- MCP write endpoints deliberately do **not** require API-key mode — they accept cookie + bearer auth (documented in `mcp.go`), so `iai login` users and the copilot both work.
- Follow-up idea (not in this PR): cross-check `create-from-catalog --auth-type` against the catalog entry's advertised `auth_methods`.


gabrielgustavoandrade (approved):

cmd/integrations.go
```diff
@@ -0,0 +1,593 @@
...
+Create a connection, verify it to discover its tools, then run a tool directly.`,
+	}
+
+	parentCmd.AddCommand(
+		makeIntegrationsListCmd(),
+		makeIntegrationsGetCmd(),
+		makeIntegrationsCatalogCmd(),
+		makeIntegrationsCreateCustomCmd(),
+		makeIntegrationsCreateFromCatalogCmd(),
+		makeIntegrationsDeleteCmd(),
```
ajaen-ai, line 42: Infrastructure group commands use `describe` as the primary name, not `get`. See `agents.go`, `databases.go`, `services.go`, `replicas.go` — all use `Use: "describe <name>"` with `Aliases: []string{"desc"}`. Flip `get`/`describe` here to match.

cmd/integrations.go
```diff
@@ -0,0 +1,593 @@
...
+)
+
+var (
+	validMcpAuthTypes  = []string{"api_key", "bearer", "none"}
+	validMcpTransports = []string{"streamable_http", "sse"}
+)
+
+func init() {
+	parentCmd := &cobra.Command{
+		Use:     "integrations",
```
ajaen-ai, line 25: "Integration" already refers to the servers that abstract agent-to-tool connections (Zendesk, Jira, etc.) in the product. Reusing it for MCP connections will confuse users. Consider `mcp` or `connections` as the primary command name.

cmd/integrations.go
```diff
@@ -0,0 +1,593 @@
...
+		description     string
+		headers         []string
+		project         string
+		org             string
+	)
+	cmd := &cobra.Command{
+		Use:   "create-custom <connection_name>",
+		Short: "Connect a custom MCP endpoint",
+		Long: `Create an integration connection to a custom (user-defined) MCP endpoint.
+
```
ajaen-ai, line 214: Every other command in the CLI uses a single `create` subcommand — `agents create`, `databases create`, `services create`, etc. None use hyphenated variants.

Merge `create-custom` and `create-from-catalog` into one `create` command. When `--catalog-id` is provided it is a catalog connection; otherwise it is custom (require `--endpoint-url` in that case).

internal/clients/mcp_test.go
```diff
@@ -0,0 +1,275 @@
+package clients
+
+import (
+	"context"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
```
ajaen-ai, line 10: Client tests here are built on `httptest` fakes. Our convention is to use real implementations in tests, not httptest mocks.

cmd/integrations.go
```diff
@@ -0,0 +1,593 @@
...
+Examples:
+  iai integrations list`,
+		Args: cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			out := cmd.OutOrStdout()
+			pCtx, apiClient, _, err := resolveProject(cmd.Context(), org, project)
+			if err != nil {
+				return err
+			}
+			data, err := apiClient.ListMcpConnections(cmd.Context(), pCtx.orgId, pCtx.projectId)
```
ajaen-ai, line 68: Every other command file declares project/org as package-level vars and reuses them across subcommands (e.g. `agentProject`, `agentOrganization` in `agents.go`). Use the same pattern here instead of redeclaring local vars in each factory function.

internal/output/integrations.go
```diff
@@ -0,0 +1,143 @@
...
+	for i, c := range conns {
+		rows[i] = []string{
+			c.ID,
+			c.Name,
+			c.Type,
+			c.Status,
+			fmt.Sprintf("%d", c.ToolCount),
+			c.EndpointURL,
+			LocalTime(c.UpdatedAt),
+		}
```
ajaen-ai, line 28: `HasCredential` is never displayed. For connections that require auth, users have no way to confirm a credential is set from CLI output.

cmd/integrations.go
```diff
@@ -0,0 +1,593 @@
...
+}
+
+func makeIntegrationsToolsRunCmd() *cobra.Command {
+	var (
+		argsJSON string
+		argsFile string
+		project  string
+		org      string
+	)
+	cmd := &cobra.Command{
```
ajaen-ai, line 539: The `tools` sub-group exists only to hold `run`. A single-child subgroup adds unnecessary nesting (`iai integrations tools run` vs `iai integrations run-tool`). Either add the other planned tools subcommands now or flatten until the group is justified.

ajaen-ai (changes requested):
Several convention breaks need to be addressed before merging. See inline comments.

cmd/integrations.go
```diff
@@ -0,0 +1,593 @@
...
+	parentCmd := &cobra.Command{
+		Use:     "integrations",
+		Aliases: []string{"integration", "mcp"},
+		Short:   "MCP integration connections for a project",
+		GroupID: groupInfra,
+		Long: `Manage Model Context Protocol (MCP) integration connections in an InteractiveAI project.
+
+Connections let agents reach external tools exposed by an MCP server — either a
+curated catalog entry (a vendor-hosted server) or a custom endpoint you define.
+Create a connection, verify it to discover its tools, then run a tool directly.`,
```
ajaen-ai, line 33: Suggested rewording for the `Long` description:

```
Manage MCP connections in a project.

Connections store the endpoint, transport, and credentials for an MCP server.
Pick one from the platform catalog or define a custom endpoint. Once connected,
verify to discover available tools and run them directly.
```

Why:
- Drops "Model Context Protocol" — anyone using this already knows what MCP stands for, and the full name adds nothing.
- Drops "InteractiveAI" — redundant in a CLI called `iai`.
- Describes what a connection actually holds (endpoint, transport, credentials) instead of the abstract "integration connections" phrasing.
- Keeps the catalog vs custom distinction without overexplaining it.
- Trims from 4 lines to 3.

ajaen-ai: Same cleanup should apply to all `Short` and `Long` descriptions in the file. Drop "integration connection" → "connection", drop "Model Context Protocol", cut filler.

A few examples:

`list` Long: `"List MCP connections in a project, showing type, status, tool count, and endpoint."`

`delete` Long: `"Delete a connection and its cached tools. Does not affect the remote MCP server. Use -f to skip confirmation."`

`tools run` Long: `"Invoke a tool on a connection and print the result. Only enabled tools can run. Pass arguments as a JSON object with --args or --args-file; omit both to send an empty object."`

cmd/integrations.go
```diff
@@ -0,0 +1,593 @@
...
+	if authType == "none" && credential != "" {
+		return fmt.Errorf("--credential must not be set when --auth-type is 'none'")
+	}
+	if authType != "none" && credential == "" {
+		return fmt.Errorf("--credential is required when --auth-type is %q", authType)
+	}
+	return nil
+}
+
+func validateMcpTransport(transport string) error {
```
ajaen-ai, line 161: Validation and input parsing functions (`validateMcpAuth`, `validateMcpTransport`, `parseHeaderFlags`, `resolveCredential`, `resolveToolArgs`) should live in `internal/inputs/mcp.go`, not in `cmd/`. That's where the rest of the codebase puts this kind of logic — see `ValidateSecretValue` in `inputs/secrets.go`, `BuildAgentRequestBody` in `inputs/agents.go`, etc.

`confirmDeletion` can stay in `cmd/` since it's interactive I/O.

internal/clients/mcp.go
```diff
@@ -0,0 +1,189 @@
+package clients
+
+import (
+	"context"
+	"net/url"
+)
```
ajaen-ai, line 6: All `APIClient` methods live in `api_client.go` or, for larger subsets, follow the `api_client_eval.go` convention. This file should be `api_client_mcp.go` to match that pattern — not a standalone `mcp.go`.

cmd/integrations_test.go
```diff
@@ -0,0 +1,244 @@
...
+	if err := validateMcpTransport("grpc"); err == nil {
+		t.Fatal("expected error for invalid transport")
+	}
+}
+
+func TestParseHeaderFlags(t *testing.T) {
+	got, err := parseHeaderFlags([]string{"X-A=1", "X-B=two=2"})
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
```
ajaen-ai, line 139: All tests in the repo use table-driven structs with `t.Run(tt.name, ...)`. Some tests in this PR follow that (`TestConfirmDeletion`, `TestValidateMcpAuth`) but several don't — `TestValidateMcpTransport`, `TestParseHeaderFlags`, `TestResolveToolArgs`, `TestResolveCredential`, `TestResolveToolArgsFromFile` are all inline assertions. Same issue in `integrations_test.go` in `output/`. Refactor these to match the table-driven pattern used everywhere else (see `secrets_test.go`, `databases_test.go`, `skills_test.go` for examples).

PR #140, opened 2026-06-02
refactor(DEV-742): address owner review on MCP connector commands

ppalacio-iai:
Follow-up to #135 (the MCP connection commands merged before review), working through @ajaen-ai's review comments. Linear: [DEV-742](https://linear.app/interactive-ai/issue/DEV-742/connectors-cli-review).

## Review feedback addressed
- **`integrations` → `connectors`** — hard rename, no back-compat alias (the product already overloads "Integration" for Zendesk/Jira servers). `cmd/integrations.go`, `output/integrations.go`, tests and docs renamed to match.
- **`get` → `describe`** as the primary verb (`get`/`desc` aliases), matching `agents`/`databases`/`services`.
- **Merged `create-custom` + `create-from-catalog` into one `create`** — `--catalog-id` ⇒ catalog connector, otherwise custom with required `--endpoint-url`; mutually-exclusive flag guards enforce it. This reworks the catalog `endpoint_url` forwarding from #139 into the unified command.
- **Validation/parsing moved to `internal/inputs/mcp.go`** (`ValidateMcpAuth`, `ValidateMcpTransport`, `ParseHeaderFlags`, `ResolveCredential`, `ResolveToolArgs`, `CatalogEndpointURL`). `confirmDeletion` stays in `cmd` (interactive I/O).
- **Client `mcp.go` → `api_client_mcp.go`** to match the `api_client_*.go` convention.
- **Factory functions → package-level command vars** with shared `connectorProject`/`connectorOrganization`.
- **Flattened `tools run` → `run-tool`** (dropped the single-child subgroup).
- **`HasCredential` now displayed** ("Credential Set:") for auth-requiring connectors.
- **Reworded** Short/Long help throughout; **tests converted to table-driven**.

## Bug fixes (from the automated review)
- `run-tool` now **returns a non-zero exit code on failed tool calls** (was always exit 0 — broke `&&` chaining).
- `VerifyMcpConnection` sends `{}` instead of `nil` so **`Content-Type: application/json` is set** (test asserts both).
- `McpToolCallData.Result` → `json.RawMessage` (preserves array/scalar results); `CustomHeaders` gets `omitempty`.

## One item I pushed back on
The "no `httptest` mocks" comment on `mcp_test.go`: the canonical client test (`internal/clients/api_client_test.go`) and `cmd/port_forward_test.go` both use `httptest.NewServer` — it's the established pattern for the HTTP client layer, with nothing to swap to. I kept `httptest` and instead converted the inline assertions to table-driven. @ajaen-ai — if you meant a specific helper, happy to adjust.

## Verification
`go build ./...`, `go vet`, and `go test ./...` all pass. Docs regenerated via `make docs`.


cmd/connectors.go
```diff
@@ -0,0 +1,430 @@
...
+	connectorCreateCmd.Flags().
+		StringVar(&connectorSlug, "slug", "", "Tool prefix used as <slug>:<tool> (auto-derived from name if omitted)")
+	connectorCreateCmd.Flags().StringVar(&connectorDescription, "description", "", "Human-readable description")
+	connectorCreateCmd.Flags().
+		StringArrayVar(&connectorHeaders, "header", nil, "Extra header as KEY=VALUE for a custom connector (repeatable)")
+	connectorCreateCmd.Flags().
+		StringVarP(&connectorProject, "project", "p", "", "Project name that owns the connector")
+	connectorCreateCmd.Flags().
+		StringVarP(&connectorOrganization, "organization", "o", "", "Organization name that owns the project")
+	connectorCreateCmd.MarkFlagsMutuallyExclusive("credential", "credential-stdin")
```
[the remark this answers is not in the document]
ajaen-ai: I agree with claude here @ppalacio-iai

cmd/connectors.go
```diff
@@ -0,0 +1,434 @@
...
+		}
+		res, err := apiClient.RunMcpTool(cmd.Context(), pCtx.orgId, pCtx.projectId, id, tool, toolArgs)
+		if err != nil {
+			return err
+		}
+		return emitToolResult(out, res)
+	},
+}
+
+// emitToolResult prints a successful tool call and, for any non-ok status,
```
ajaen-ai: I think this should stay in `cmd`, not `internal/output`. The `output` package should handle formatting/printing, but this function also decides command behavior: non-`ok` tool results become returned errors so Cobra exits non-zero. The formatting is already delegated to `output.McpToolCallError` / `PrintMcpToolResult`.

The code comment could be shorter though, e.g.:

```go
// emitToolResult returns failed tool calls as errors so the command exits non-zero.
```

docs/iai_connectors.md
```diff
@@ -0,0 +1,38 @@
+## iai connectors
+
+Manage MCP connectors in a project
+
+### Synopsis
+
+Manage MCP connectors in a project.
```
ajaen-ai: Small docs nit: this repetition comes from Cobra generating the command `Short` before the Synopsis, then using `Long` as the Synopsis body. Since `Long` starts with the same sentence as `Short`, the generated docs repeat themselves.

I’d avoid that by starting `Long` with the more detailed explanation instead. That keeps the short description as the one-line summary and the synopsis as the extra context.

internal/clients/api_client_mcp_test.go
```diff
@@ -194,13 +200,17 @@ func TestDeleteMcpConnection(t *testing.T) {
...
 func TestVerifyMcpConnection(t *testing.T) {
+	var gotBody, gotContentType string
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if r.Method != http.MethodPost {
 			t.Fatalf("method = %s, want POST", r.Method)
 		}
 		if r.URL.Path != "/api/platform/v1/organizations/org-1/projects/proj-1/mcp-connections/c1/verify" {
 			t.Fatalf("path = %s", r.URL.Path)
 		}
+		gotContentType = r.Header.Get("Content-Type")
```
ajaen-ai: Can we avoid using `httptest` here? The repo guidelines prefer pure unit tests without fake HTTP servers/mocks. For this PR, I think we can cover most of the behavior with table-driven tests around request/body construction, parsing, and output formatting, and leave real API behavior to integration tests.

internal/inputs/mcp.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"github.com/Interactive-AI-Labs/interactive-cli/internal/clients"
+)
+
+var (
+	validMcpAuthTypes  = []string{"api_key", "bearer", "none"}
+	validMcpTransports = []string{"streamable_http", "sse"}
+)
+
+func ValidateMcpAuth(authType, credential string) error {
```
ajaen-ai: Can we reduce the validation here? We usually keep `internal/inputs` for CLI-only input handling/parsing or special cases the server cannot validate, and otherwise let the backend return the authoritative error message.

Helpers like reading `--credential-stdin`, parsing `--args-file`, and converting `--header KEY=VALUE` are CLI input handling, so they make sense. But `ValidateMcpAuth` / `ValidateMcpTransport` duplicate backend validation/enums. I’d prefer removing those unless there’s a specific reason the server cannot return a useful error for these cases.

cmd/connectors.go
```diff
@@ -0,0 +1,474 @@
...
+		out := cmd.OutOrStdout()
+		pCtx, apiClient, _, err := resolveProject(
+			cmd.Context(),
+			connectorOrganization,
+			connectorProject,
+		)
+		if err != nil {
+			return err
+		}
+		data, err := apiClient.ListMcpConnections(cmd.Context(), pCtx.orgId, pCtx.projectId)
```
ajaen-ai: Should this support the standard structured output flags? For new resource commands we usually want machine-readable output as well as tables. Can we add `--json` / `--yaml` where applicable, at least for `connectors list`, `catalog`, `describe`, and `verify`, so agents don’t have to parse table output?

cmd/connectors.go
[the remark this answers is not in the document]
ppalacio-iai: Done in bf68c8b — moved `--project`/`--organization` to `connectorsCmd.PersistentFlags()`, so the 12 per-subcommand registrations collapse to 2. No behaviour change; every subcommand still inherits both.

cmd/connectors.go
ppalacio-iai: Agreed — kept it in `cmd` (it decides exit behaviour; the formatting is delegated to `output.McpToolCallError` / `PrintMcpToolResult`). Adopted your exact one-liner for the comment in bf68c8b.

docs/iai_connectors.md
ppalacio-iai: Fixed in bf68c8b — reworded `Long` on the connector commands to lead with the detailed explanation instead of repeating `Short`, and regenerated the docs. The synopsis no longer duplicates the short description.

internal/clients/api_client_mcp_test.go
ppalacio-iai: Done in bf68c8b — `api_client_mcp_test.go` no longer uses `httptest`. It is now table-driven: request-body construction via `json.Marshal` (catalog vs custom, omitempty rules) and response parsing through the real `decodeSuccess` path (including the `json.RawMessage` object/array/scalar result shapes).

One trade-off worth flagging: wire-only behaviour that needs a live server — URL paths, the `Content-Type: application/json` on verify, and the nil-args -> `{}` defaulting in `RunMcpTool` — is no longer unit-covered and now relies on integration tests, as you suggested.

For context, `httptest.NewServer` is still used by `internal/clients/api_client_test.go` and `cmd/port_forward_test.go`. If you want the no-`httptest` rule applied across the whole client layer (or written into AGENTS.md section 3), happy to open a follow-up.

internal/inputs/mcp.go
ppalacio-iai: Done in bf68c8b — removed `ValidateMcpAuth` and `ValidateMcpTransport` (they duplicated the backend enums), so auth/transport errors now come from the platform. Kept the CLI-only helpers you called out: `ResolveCredential` (--credential-stdin), `ResolveToolArgs` (--args-file), and `ParseHeaderFlags` (--header KEY=VALUE). `--auth-type` stays `MarkFlagRequired`, since that is a CLI-level concern.

cmd/connectors.go
ppalacio-iai: Added `--json` to `list`, `catalog`, `get`, and `verify` in bf68c8b (raw API response, via the shared `PrintRawJSON` + raw-body capture that the other resource commands use).

I left out `--yaml`: the standard output flag in AGENTS.md section 5 is `--json` only, and no command in the repo exposes `--yaml` today. Happy to add it if you want it as a new convention, but that would be a repo-wide change rather than connector-specific.

internal/clients/api_client_mcp_test.go
ajaen-ai: Yes. I will soon add integration tests in the same way I did for the platform backend.

Thanks for flagging it.

cmd/connectors.go
ajaen-ai: I have just added yaml in another PR so if you can add it that would be great. 

Can you also update the AGENTS.md?

ppalacio-iai: Done in d2e640d — merged `main` to pick up #141, then added `--yaml` alongside `--json` on `list`, `get`, `catalog`, and `verify`, rendering the raw payload via `output.PrintRawYAML` (the same helper #141 introduced). Also documented `--yaml` as a standard flag in AGENTS.md §5. Docs regenerated.

ajaen-ai (approved):

PR #141, opened 2026-06-04
DEV-747: Add structured json and yaml output

ajaen-ai:


ajaen-ai:
Fixed the review items.

- Updated YAML mode to bypass column validation in traces, comments, metrics, run-items, and dataset-items.
- Moved PrintRawJSON next to PrintRawYAML in internal/output/structured.go.
- Fixed YAML numeric normalization so numbers are emitted as numbers instead of quoted strings.
- Added table-driven tests for raw/structured output and representative --json/--yaml mutual exclusivity.
- Revalidated with read-only `go run main.go` commands and `go test ./...`.

@claude can you review again?

ajaen-ai:
Updated again.

- `schema --json` now emits valid structured JSON (`schema` + `schemaVersion`) instead of a text header plus raw schema.
- Added `--columns` conflict validation so table-only columns cannot be combined with `--json`/`--yaml`.
- Centralized compatibility matrix JSON output through `PrintStructuredJSON`.
- Re-ran docs generation and `go test ./...`; pre-commit hooks passed.

@claude can you review again?

ajaen-ai:
Removed stale generated `describe` docs after the primary command rename to `get`; kept unrelated legacy docs that are not generated by `gen-docs`. Re-ran `go test ./...` and pushed the amended commit.

@claude can you review again?

PR #146, opened 2026-06-16
docs(cli): add missing command examples

dgallego-iai:
Refs DEV-786.

## What
Scan all CLI commands and add usage examples to the leaf commands that were missing them (e.g. `service create`, the example called out in the issue).

While doing so, standardize on Cobra's native `Example:` field instead of embedding examples inside the `Long:` description. Cobra renders `Long` into the markdown "Synopsis" as a plain paragraph, which collapsed the indentation/newlines of the example block (it looked fine in `--help` but wrapped/ran together in the generated docs). The `Example:` field renders inside a fenced code block, so both `--help` and the markdown look correct.

## Changes
- Add examples to every leaf command that lacked them.
- Migrate **all** example blocks (new and pre-existing) from `Long:` into the native `Example:` field — including the templated prompt-type system (`routines`/`policies`/`variables`/`glossaries`/`macros`, via new `*Example` fields on `PromptTypeConfig`) and `skills`.
- Regenerate `docs/`.

## Verification
- `gofmt` clean, `go build ./...` passes, `go test ./...` passes.
- No `Examples:` blocks remain embedded in any `Long:`.

### DIFF
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -1,5 +1,5 @@
 [tool.bumpversion]
-current_version = "0.34.3"
+current_version = "0.34.4"
 commit = true
 tag = false
 parse = "(?P<major>\\d+)\\.(?P<minor>\\d+)\\.(?P<patch>\\d+)(-beta\\.(?P<dev>\\d+))?"
--- a/cmd/agents.go
+++ b/cmd/agents.go
@@ -73,10 +73,8 @@ The config schema depends on the agent version. Run
 
 Routines and policies referenced in the config must already exist in the project
 and should be validated against the matching schema version (see --schema-version
-on their create/update commands).
-
-Examples:
-  iai agents create chat-agent --id interactive-agent --version 0.0.1 --file agent-config.yaml
+on their create/update commands).`,
+	Example: `  iai agents create chat-agent --id interactive-agent --version 0.0.1 --file agent-config.yaml
   iai agents create chat-agent --id interactive-agent --version 0.0.1 --file agent-config.yaml --endpoint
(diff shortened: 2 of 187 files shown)


dgallego-iai:
Addressed in a follow-up commit (`fix(docs): preserve CLI examples in prompt-type create/update docs`), kept separate from the examples commit.

**🔴 Incomplete docs for prompt-type create/update — fixed.**
Root cause was in `gen_docs.go`, not stale docs (which is why `make docs` + the `git diff --exit-code docs/` hook both passed). `injectSchemaDoc` replaced everything from the `Run 'iai … schema'` marker through to `### Options`, which swallowed the cobra-generated `### Examples` (CLI) block and left only the schema template's `### Example` (YAML). This was actually true on `main` too — those CLI examples never reached the docs.

Fix: stop the replacement at `### Examples` when present (falling back to `### Options`). The 10 affected docs (`routines`/`policies`/`variables`/`glossaries`/`macros` × create/update) now render **both** the schema `### Example` and the CLI `### Examples`. `skills` create/update were already correct — they aren't schema-injected — so no change was needed there.

**🟡 File-format example remains in `Long:` — by design, and moot in these rendered docs.**
The `Example (<file>):` block stays in `Long:` intentionally (it illustrates file content, not a CLI invocation — same call as the `skills.md` block). For create/update it sits *after* the schema marker, so the schema template replaces it in the generated docs anyway; it only appears in terminal `--help`, where whitespace is preserved, so the collapse concern doesn't surface.

All pre-commit hooks and CI pass.

ajaen-ai (approved):

PR #147, opened 2026-06-19
Fix log timestamps and shorten replica IDs

ajaen-ai:
## Summary
- restore timestamps in formatted log output
- convert Unix nanosecond log timestamps to local time
- shorten replica prefixes to the suffix after the last hyphen

## Tests
- go test ./internal/output ./cmd

### DIFF
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -1,5 +1,5 @@
 [tool.bumpversion]
-current_version = "0.34.4"
+current_version = "0.34.5"
 commit = true
 tag = false
 parse = "(?P<major>\\d+)\\.(?P<minor>\\d+)\\.(?P<patch>\\d+)(-beta\\.(?P<dev>\\d+))?"
--- a/cmd/agents.go
+++ b/cmd/agents.go
@@ -379,14 +379,15 @@ var agentRestartCmd = &cobra.Command{
 }
 
 var (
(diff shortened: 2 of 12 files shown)


ajaen-ai:
Updated based on the review and follow-up discussion:

- removed inner timestamp fallback/handling; log entries always rely on the outer timestamp
- kept numeric timestamp formatting log-only, not in shared LocalTime
- kept replica color keys on the full replica name while displaying only the suffix
- simplified tests to cover the actual outer timestamp path

@claude please re-review when you get a chance.

ajaen-ai:
Updated again: timestamps now follow kubectl-style behavior.

- default log output does not print the platform/Loki timestamp
- new --timestamps flag prints the outer platform timestamp
- replica prefix shortening still applies by default
- raw output remains untouched

Tests/hooks passed.

ajaen-ai:
Updated after the last review: timestamp display now matches kubectl-style behavior with an explicit --timestamps flag, while replica suffix shortening remains enabled by default.

@claude please re-review the latest version.

ajaen-ai:
Added --timestamps examples to the log command docs. Leaving --raw + --timestamps allowed intentionally: raw output already includes the timestamp field, and raw should keep server output untouched.

@claude thanks for the review — examples are updated.

PR #151, opened 2026-06-26
DEV-813: add CLI key management commands

ajaen-ai:
## Summary
- add project API key list/create/delete commands
- add router API key list/create/delete commands
- require login-session cookie auth for key management
- consolidate shared output formatting helpers and show router limits as USD

## Test plan
- go test ./...
- go run main.go api-keys list/create/delete
- go run main.go router-keys list/create/delete
- go run main.go --api-key pk:sk router-keys list

### DIFF
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -1,5 +1,5 @@
 [tool.bumpversion]
-current_version = "0.34.6"
+current_version = "0.35.0"
 commit = true
 tag = false
 parse = "(?P<major>\\d+)\\.(?P<minor>\\d+)\\.(?P<patch>\\d+)(-beta\\.(?P<dev>\\d+))?"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -48,7 +48,7 @@ repos:
 
       - id: run-unit-tests
         name: Run unit tests
-        entry: go test ./...
+        entry: env TZ=Europe/Berlin go test ./...
         language: system
         pass_filenames: false
         always_run: false
--- a/cmd/api_keys.go
+++ b/cmd/api_keys.go
@@ -1,0 +1,403 @@
+package cmd
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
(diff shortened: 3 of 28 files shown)


ajaen-ai:
@claude please rereview.

Addressed:
- login-only guard now checks that an iai login session exists before continuing
- project API key delete no longer hardcodes a response message
- --expires-at now validates RFC3339 client-side
- list commands now support --columns with minimal defaults and optional extra fields
- router key flag docs include backend defaults without environment wording

Intentionally kept RouterAPIKey limit fields as flexible values because the backend returns mixed string/number JSON for limits.

ajaen-ai:
@claude please rereview.

Addressed the remaining items:
- added --limit-reset requires --limit validation
- delete commands now check res.Success before printing success
- list --columns is now mutually exclusive with --json/--yaml
- kept RouterAPIKey limit fields flexible because backend returns mixed string/number JSON

Also set CI TZ to Europe/Berlin so existing timezone-dependent tests match local expectations.

ajaen-ai:
@claude please rereview.

Latest updates:
- moved the shared key-management login-required error into internal/auth and reused it from cmd/client code
- kept the command-local session check in cmd because it depends on cmd globals and session files
- updated router key help text to match deployed defaults without mentioning environments: limit defaults to 00, max 500, reset defaults to monthly, omitted expiration means no expiry
- added explicit --limit-reset requires --limit validation
- delete commands check res.Success before printing success
- --columns is mutually exclusive with --json/--yaml

ajaen-ai:
Follow-up fixes pushed:

- removed the dead `res.Message` success-print branch from project API key delete; project deletes now always print `Deleted` after `Success=true`
- kept router delete message handling because that response is server-unmarshalled and can include a message
- centralized the test timezone in `.pre-commit-config.yaml` with `env TZ=Europe/Berlin go test ./...`
- removed the workflow-level `TZ` so local and GitHub Actions use the same pre-commit behavior

Also manually tested the full command matrix: help, aliases, table/json/yaml, columns, validation failures, auth rejection, server errors, real create/delete, and cleanup verification.

PR #154, opened 2026-06-30
feat(collections): iai collections CLI commands

dgallego-iai:
## Summary

Adds `iai collections` — CLI coverage for the knowledge-base / pgvector **collections** endpoints in the deployment-operator (paired with deployment-operator#186). A collection lives inside a database, so every command is scoped by a required `-d/--database` (mirroring how `dataset-items`/`replicas` name their parent), consistent with how each CLI resource is named after its API resource.

Naming decision: **`iai collections`** (1:1 with the API resource), chosen over `iai knowledgebase` / nesting under `databases` (see the design doc; validated with the team).

## Command surface

Built in five vertical slices, each its own commit:

- **lifecycle** — `list` / `create --file` / `describe` / `patch --file` / `delete` / `stats`
- **chunks** — `upsert --file` / `list` / `get [--include-vector]` / `count [--filter]` / `patch --file` / `delete` / `bulk-delete (--ids|--filter|--all)`
- **documents** — `list` / `get` / `delete`
- **search** — `search --query|--vector [--exact]` / `search batch --file` / `search by-id --id` / `search hybrid --file`
- **slots** — `slots add` / `reindex` / `vacuum` / `progress` / `delete`

Same 4-layer wiring as the other resources: `internal/clients` (DeploymentClient methods), `internal/inputs` (request bodies; YAML or JSON via `--file`), `internal/output` (table formatters + `--json`/`--yaml`), `cmd`. Regenerated CLI docs via `make docs`.

## Review-fix pass

The Claude bot review (10 issues: 1 red, 4 yellow, 3 green, 2 nits) was addressed in 9 follow-up commits on top of the original 15. #9 (`--json`/`--yaml` on `create`/`patch`) was deliberately skipped — it matches the established mutation-command pattern across the CLI (no `--json`/`--yaml` on `agentCreateCmd`, `databaseCreateCmd`, `collDeleteCmd`, etc.).

| # | Commit | Subject |
|---|---|---|
| 1 (red) | `1cd06dd` | expose DeletedIds in chunks bulk-delete output |
| 2 (yel) | `71681c5` | inline GET for slot index-progress |
| 3 (yel) | `12ad8ac` | surface real I/O errors in bulk-delete confirm prompt |
| 4 (yel) | `26b74e0` | add --json/--yaml to slots reindex and vacuum |
| 5 (yel) | `d03dea5` | drop dead empty-path check in ReadCollectionBodyFile |
| 6 (grn) | `2a01cb4` | fold body-read error into collectionErr fallback |
| 7 (grn) | `c5c9301` | document --type valid values and default |
| 8 (grn) | `549f373` | cover every output formatter |
| 9 (nit) | _skipped_ | matches established mutation-command pattern |
| 10 (nit) | `1507073` | sort byStatus keys in PrintChunkUpsertResult |

## Test plan

`go build ./...`, `go vet ./...`, `go test ./...`, golines/gofumpt/gci, `make docs` — all green (pre-commit clean).

**Manual end-to-end against a live deployment-operator** (local `air`, dev pgvector DB): 69 steps across two scripts, **0 failures**.

### Smoke (`scripts/e2e-collections.sh` — 36/36)
- One fresh `e2e-<ts>` collection per run, full lifecycle pass.
- Read smoke on the existing `docs` collection.
- Create with `embedding` shorthand + `full_text` + a `extra` 4-d raw slot.
- Slot lifecycle (progress / add / vacuum / delete).
- Chunk CRUD (upsert / list / get / count / patch / bulk-delete).
- Document CRUD (list / get / delete).
- Search matrix (query / vector / exact / by-id / batch / hybrid).
- Patch collection, slot reindex, final stats, delete collection.

### Negatives + output modes (`scripts/e2e-collections-neg.sh` — 33/33)
- **CLI-side validation (7):** `bulk-delete` no/multi selector, search no/dual input, `by-id` missing id, `count` bad JSON, `create` no file / bad YAML.
- **Server-side validation (3):** `slot add` with `dimension: 0` / missing, `patch` changing immutable field, `chunks upsert` with vector dim mismatch (returns 207 with per-chunk errors — verified via `--json`).
- **404 / not-found (12):** describe/stats on missing collection, chunk get/delete/patch on missing chunk, document get/delete, slot progress/delete/reindex/vacuum on missing slot, chunk route through missing collection.
- **Confirmation prompt (1):** `bulk-delete --all` with `n` on stdin → aborts cleanly.
- **Output modes (9):** `--json` shape for `collections list/describe`, `chunks get`, `chunks bulk-delete`, `documents list`, `slots progress/reindex/vacuum`.

Runbooks and scripts at `~/git/iai-personal-knowledge-base/docs/processes/collections-e2e.md` and `scripts/e2e-collections*.sh`.

## Surfaced by the test

- `collections list --json` returns a raw array `[...]`, not a wrapped object — verified by the `json_collections_list` shape check.
- Vector dim mismatch on `chunks upsert` returns 207 with per-chunk errors, not a hard 4xx — verified by `neg_chunks_dim_mismatch`.

Refs: DEV-818

### DIFF
--- a/cmd/collections.go
+++ b/cmd/collections.go
@@ -1,0 +1,411 @@
+package cmd
+
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/Interactive-AI-Labs/interactive-cli/internal/clients"
+	"github.com/Interactive-AI-Labs/interactive-cli/internal/inputs"
+	"github.com/Interactive-AI-Labs/interactive-cli/internal/output"
+	"github.com/spf13/cobra"
+)
+
+var (
+	collProject      string
+	collOrganization string
+	collDatabase     string
+	collFile         string
+	collJSON         bool
+	collYAML         bool
+	collDryRun       bool
+)
+
+var collectionsCmd = &cobra.Command{
+	Use:     "collections",
+	Aliases: []string{"collection", "coll"},
+	Short:   "Knowledge bases (searchable tables of chunks) inside a pgvector database",
+	GroupID: groupInfra,
+	Long: `Manage collections within a database.
+
+A collection is a table of chunks (rows) — each chunk is text plus its vector
+embedding(s) — that you search by meaning or keyword; it's what backs a
+knowledge base. It lives inside an existing pgvector database, so every command
+requires --database. Use 'iai databases create' first to provision the database.
+
+Run 'iai collections schema' to see the body format for every --file command.`,
+}
+
+var collListCmd = &cobra.Command{
+	Use:     "list",
+	Aliases: []string{"ls"},
+	Short:   "List collections in a database",
+	Example: `  iai collections list -d my-db
+  iai collections list -d my-db --json`,
+	Args: cobra.NoArgs,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		out := cmd.OutOrStdout()
+
+		pCtx, _, deployClient, err := resolveProject(cmd.Context(), collOrganization, collProject)
+		if err != nil {
+			return err
+		}
+
+		collections, err := deployClient.ListCollections(
+			cmd.Context(),
+			pCtx.orgId,
+			pCtx.projectId,
+			collDatabase,
+		)
+		if err != nil {
+			return err
+		}
+
+		if collJSON {
+			return output.PrintStructuredJSON(out, collections)
+		}
+		if collYAML {
+			return output.PrintStructuredYAML(out, collections)
+		}
+		return output.PrintCollectionList(out, collections)
+	},
+}
+
+var collDescribeCmd = &cobra.Command{
+	Use:     "describe <collection>",
+	Aliases: []string{"desc"},
+	Short:   "Describe a collection's configuration",
+	Example: `  iai collections describe docs -d my-db
+  iai collections describe docs -d my-db --json`,
+	Args: cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		out := cmd.OutOrStdout()
+		name := strings.TrimSpace(args[0])
+
+		pCtx, _, deployClient, err := resolveProject(cmd.Context(), collOrganization, collProject)
+		if err != nil {
+			return err
+		}
+
+		coll, err := deployClient.DescribeCollection(
+			cmd.Context(),
+			pCtx.orgId,
+			pCtx.projectId,
+			collDatabase,
+			name,
+		)
+		if err != nil {
+			return err
+		}
+
+		if collJSON {
+			return output.PrintStructuredJSON(out, coll)
+		}
+		if collYAML {
+			return output.PrintStructuredYAML(out, coll)
+		}
+		return output.PrintCollectionDescribe(out, coll)
+	},
+}
+
+var collStatsCmd = &cobra.Command{
+	Use:   "stats <collection>",
+	Short: "Show a collection's chunk count, size, and index status",
+	Example: `  iai collections stats docs -d my-db
+  iai collections stats docs -d my-db --json`,
+	Args: cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		out := cmd.OutOrStdout()
+		name := strings.TrimSpace(args[0])
+
+		pCtx, _, deployClient, err := resolveProject(cmd.Context(), collOrganization, collProject)
+		if err != nil {
+			return err
+		}
+
+		stats, err := deployClient.GetCollectionStats(
+			cmd.Context(),
+			pCtx.orgId,
+			pCtx.projectId,
+			collDatabase,
+			name,
+		)
+		if err != nil {
+			return err
+		}
+
+		if collJSON {
+			return output.PrintStructuredJSON(out, stats)
+		}
+		if collYAML {
+			return output.PrintStructuredYAML(out, stats)
+		}
+		return output.PrintCollectionStats(out, stats)
+	},
+}
+
+var collCreateCmd = &cobra.Command{
+	Use:   "create <collection>",
+	Short: "Create a collection from a config file",
+	Long: `Create a vector collection from a YAML or JSON config file (--file).
+
+The config declares the vector slot(s) — either an embedding-backed slot
+("embedding": {model, dimension}) or a raw vector slot ({type, dimension,
+distance}) — and optional full-text search.
+
+Slot type, dimension, distance, and the embedding model are IMMUTABLE after
+creation; fixing a wrong value means deleting and recreating the collection.
+
+Run 'iai collections schema' for the config file format.`,
+	Example: `  iai collections create docs -d my-db --file collection.yaml`,
+	Args:    cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		out := cmd.OutOrStdout()
+		name := strings.TrimSpace(args[0])
+
+		body, err := inputs.ReadCollectionBodyFile(collFile)
+		if err != nil {
+			return err
+		}
+
+		pCtx, _, deployClient, err := resolveProject(cmd.Context(), collOrganization, collProject)
+		if err != nil {
+			return err
+		}
+
+		msg, err := deployClient.CreateCollection(
+			cmd.Context(),
+			pCtx.orgId,
+			pCtx.projectId,
+			collDatabase,
+			name,
+			body,
+			collDryRun,
+		)
+		if err != nil {
+			return err
+		}
+		if msg != "" {
+			fmt.Fprintln(out, msg)
+		}
+		return nil
+	},
+}
+
+var collPatchCmd = &cobra.Command{
+	Use:   "patch <collection>",
+	Short: "Update a collection's mutable config from a file",
+	Long: `Update a collection's mutable configuration from a YAML or JSON file (--file):
+full-text settings and per-slot ef_search_default. Slot type/dimension/distance
+and the embedding model are immutable.`,
+	Example: `  iai collections patch docs -d my-db --file patch.yaml`,
+	Args:    cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		out := cmd.OutOrStdout()
+		name := strings.TrimSpace(args[0])
+
+		body, err := inputs.ReadCollectionBodyFile(collFile)
+		if err != nil {
+			return err
+		}
+
+		pCtx, _, deployClient, err := resolveProject(cmd.Context(), collOrganization, collProject)
+		if err != nil {
+			return err
+		}
+
+		path := clients.CollectionsPath(pCtx.orgId, pCtx.projectId, collDatabase) +
+			"/" + url.PathEscape(name)
+		msg, err := deployClient.SendCollectionBody(
+			cmd.Context(), http.MethodPatch, path, body, "update collection", "",
+		)
+		if err != nil {
+			return err
+		}
+		if msg != "" {
+			fmt.Fprintln(out, msg)
+		}
+		return nil
+	},
+}
+
+var collDeleteCmd = &cobra.Command{
+	Use:     "delete <collection>",
+	Aliases: []string{"rm"},
+	Short:   "Delete a collection and all its data",
+	Example: `  iai collections delete docs -d my-db`,
+	Args:    cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		out := cmd.OutOrStdout()
+		name := strings.TrimSpace(args[0])
+
+		pCtx, _, deployClient, err := resolveProject(cmd.Context(), collOrganization, collProject)
+		if err != nil {
+			return err
+		}
+
+		msg, err := deployClient.DeleteCollection(
+			cmd.Context(),
+			pCtx.orgId,
+			pCtx.projectId,
+			collDatabase,
+			name,
+		)
+		if err != nil {
+			return err
+		}
+		if msg != "" {
+			fmt.Fprintln(out, msg)
+		}
+		return nil
+	},
+}
+
+func init() {
+	// Shared scope flags on every subcommand: -o org, -p project, -d database.
+	for _, c := range []*cobra.Command{
+		collListCmd, collDescribeCmd, collStatsCmd, collCreateCmd, collPatchCmd, collDeleteCmd,
+	} {
+		c.Flags().StringVarP(&collOrganization, "organization", "o", "", "Organization name")
+		c.Flags().StringVarP(&collProject, "project", "p", "", "Project name")
+		c.Flags().
+			StringVarP(&collDatabase, "database", "d", "", "Database that holds the collection (required)")
+		_ = c.MarkFlagRequired("database")
+	}
+
+	for _, c := range []*cobra.Command{collListCmd, collDescribeCmd, collStatsCmd} {
+		c.Flags().BoolVar(&collJSON, "json", false, "Output raw API response as JSON")
+		c.Flags().BoolVar(&collYAML, "yaml", false, "Output raw API response as YAML")
+	}
+
+	collCreateCmd.Flags().StringVar(&collFile, "file", "", "Path to a YAML/JSON collection config")
+	collCreateCmd.Flags().
+		BoolVar(&collDryRun, "dry-run", false, "Validate the config without creating the collection")
+	_ = collCreateCmd.MarkFlagRequired("file")
+	collPatchCmd.Flags().StringVar(&collFile, "file", "", "Path to a YAML/JSON patch config")
+	_ = collPatchCmd.MarkFlagRequired("file")
+
+	collectionsCmd.AddCommand(
+		collListCmd, collDescribeCmd, collStatsCmd, collCreateCmd, collPatchCmd, collDeleteCmd,
+		collSchemaCmd,
+	)
+	collSchemaCmd.Flags().BoolVar(&collJSON, "json", false, "Output the schemas as JSON")
+	collSchemaCmd.Flags().BoolVar(&collYAML, "yaml", false, "Output the schemas as YAML")
+	rootCmd.AddCommand(collectionsCmd)
+}
+
+var collSchemaCmd = &cobra.Command{
+	Use:   "schema",
+	Short: "Show the file schemas for the --file-based collection commands",
+	Long: `Print the expected shape of every --file body: collection create/patch,
+chunks upsert/patch, slots add/reindex, and search batch/hybrid. Use --json or
+--yaml for structured output.`,
+	Example: `  iai collections schema
+  iai collections schema --json`,
+	Args: cobra.NoArgs,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		out := cmd.OutOrStdout()
+		if collJSON {
+			return output.PrintStructuredJSON(out, collectionSchemas())
+		}
+		if collYAML {
+			return output.PrintStructuredYAML(out, collectionSchemas())
+		}
+		fmt.Fprint(out, collectionSchemaText)
+		return nil
+	},
+}
+
+const collectionSchemaText = `Collection file schemas (bodies for --file). Keys are snake_case
+(full_text, ef_search_default) except chunk record fields, which are camelCase (documentId).
+
+create (--file):
+  {
+    "vectors": {
+      "<slot>": { "type": "float32", "dimension": 768, "distance": "cosine" },
+      "<slot>": { "embedding": { "model": "<model-id>" }, "dimension": 768 }
+    },
+    "full_text": { "enabled": true, "language": "english" }
+  }
+  NOTE: slot type, dimension, distance, and embedding model are IMMUTABLE.
+  Provide either a raw slot (type+dimension) or an embedding slot (embedding+dimension).
+
+patch (--file):
+  { "full_text": { "enabled": false },
+    "vectors": { "<slot>": { "index": { "ef_search_default": 200 } } } }
+
+chunks upsert (--file):
+  { "chunks": [
+      { "id": "c1", "documentId": "doc1", "text": "content", "metadata": {"k":"v"} }
+  ] }
+  Fields: id, documentId, text, metadata, vector/vectors. Use documentId (NOT document_id).
+
+chunks patch (--file):
+  { "text": "new content", "metadata": {"k":"v"} }
+
+slots add / reindex (--file):
+  { "type": "float32", "dimension": 768, "distance": "cosine",
+    "index": { "type": "hnsw", "m": 16, "ef_construct": 200 } }
+
+search batch (--file):
+  { "searches": [ {"query":"text","using":"default","limit":10} ] }
+
+search hybrid (--file):
+  { "queries": [
+      {"query":"text","using":"default","candidate_limit":50},
+      {"full_text":"keywords","candidate_limit":30}
+    ],
+    "fusion": {"method":"rrf","k":60}, "limit": 10 }
+  NOTE: "using" selects a vector slot; use "full_text" for the keyword lane.
+`
+
+// collectionSchemas mirrors collectionSchemaText as a structured map for --json/--yaml.
+func collectionSchemas() map[string]any {
+	return map[string]any{
+		"create": map[string]any{
+			"vectors": map[string]any{
+				"<slot>": map[string]any{
+					"type": "float32", "dimension": 768, "distance": "cosine",
+					"embedding": map[string]string{"model": "<model-id>"},
+				},
+			},
+			"full_text": map[string]any{"enabled": true, "language": "english"},
+			"immutable": []string{"type", "dimension", "distance", "embedding.model"},
+		},
+		"patch": map[string]any{
+			"full_text": map[string]any{"enabled": false},
+			"vectors": map[string]any{
+				"<slot>": map[string]any{"index": map[string]int{"ef_search_default": 200}},
+			},
+		},
+		"chunks_upsert": map[string]any{
+			"chunks": []map[string]any{
+				{
+					"id":         "c1",
+					"documentId": "doc1",
+					"text":       "content",
+					"metadata":   map[string]string{"k": "v"},
+				},
+			},
+		},
+		"chunks_patch": map[string]any{
+			"text":     "new content",
+			"metadata": map[string]string{"k": "v"},
+		},
+		"slots_add": map[string]any{
+			"type": "float32", "dimension": 768, "distance": "cosine",
+			"index": map[string]any{"type": "hnsw", "m": 16, "ef_construct": 200},
+		},
+		"search_batch": map[string]any{
+			"searches": []map[string]any{{"query": "text", "using": "default", "limit": 10}},
+		},
+		"search_hybrid": map[string]any{
+			"queries": []map[string]any{
+				{"query": "text", "using": "default", "candidate_limit": 50},
+				{"full_text": "keywords", "candidate_limit": 30},
+			},
+			"fusion": map[string]any{"method": "rrf", "k": 60}, "limit": 10,
+		},
+	}
+}
--- a/cmd/collections_documents.go
+++ b/cmd/collections_documents.go
@@ -1,0 +1,174 @@
+package cmd
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/Interactive-AI-Labs/interactive-cli/internal/output"
+	"github.com/spf13/cobra"
+)
+
+var (
+	docLimit         int
+	docCursor        string
+	docFilter        string
+	docIncludeVector bool
+)
+
+var documentsCmd = &cobra.Command{
+	Use:     "documents",
+	Aliases: []string{"document", "docs"},
+	Short:   "Inspect documents (chunks grouped by documentId)",
+	Long:    `A document groups chunks by documentId; these commands read or delete them.`,
+}
+
+var documentsListCmd = &cobra.Command{
+	Use:     "list <collection>",
+	Aliases: []string{"ls"},
+	Short:   "List documents in a collection",
+	Example: `  iai collections documents list docs -d my-db`,
+	Args:    cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		out := cmd.OutOrStdout()
+		collection := strings.TrimSpace(args[0])
+
+		pCtx, _, deployClient, err := resolveProject(cmd.Context(), collOrganization, collProject)
+		if err != nil {
+			return err
+		}
+
+		list, err := deployClient.ListDocuments(
+			cmd.Context(),
+			pCtx.orgId,
+			pCtx.projectId,
+			collDatabase,
+			collection,
+			docLimit,
+			docCursor,
+			docFilter,
+		)
+		if err != nil {
+			return err
+		}
+
+		if collJSON {
+			return output.PrintStructuredJSON(out, list)
+		}
+		if collYAML {
+			return output.PrintStructuredYAML(out, list)
+		}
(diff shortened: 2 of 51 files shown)


dgallego-iai:
Addressed the Claude bot review from the re-opened PR (10 issues, 1 red, 4 yellow, 3 green, 2 nits). 9 fixes landed as separate commits on top of the original 15, plus this PR's regen of CLI docs; #9 (`--json`/`--yaml` on `create`/`patch`) was deliberately skipped because it matches the established mutation-command pattern across the CLI (no `--json`/`--yaml` on `agentCreateCmd`, `databaseCreateCmd`, `collDeleteCmd`, etc.).

| # | Commit | Subject |
|---|---|---|
| 1 (red) | `1cd06dd` | expose DeletedIds in chunks bulk-delete output |
| 2 (yel) | `71681c5` | inline GET for slot index-progress |
| 3 (yel) | `12ad8ac` | surface real I/O errors in bulk-delete confirm prompt |
| 4 (yel) | `26b74e0` | add --json/--yaml to slots reindex and vacuum |
| 5 (yel) | `d03dea5` | drop dead empty-path check in ReadCollectionBodyFile |
| 6 (grn) | `2a01cb4` | fold body-read error into collectionErr fallback |
| 7 (grn) | `c5c9301` | document --type valid values and default |
| 8 (grn) | `549f373` | cover every output formatter |
| 9 (nit) | _skipped_ | matches established mutation-command pattern |
| 10 (nit) | `1507073` | sort byStatus keys in PrintChunkUpsertResult |

`go build ./...`, `go vet ./...`, `go test ./...` all green.

**End-to-end verification** against the live `deployment-operator` on `localhost:8000` with a fresh `e2e-<ts>` collection: 36/36 PASS — every new command exercised (lifecycle, chunks, documents, search, slots). Runbook at `~/git/iai-personal-knowledge-base/docs/processes/collections-e2e.md`, script at `scripts/e2e-collections.sh`.

Ready for review.

interactivealex:
Ran the collections CLI end to end migrating a real KB (new DB → config → 128 chunks → full-text → search). Stuff that tripped me up, grouped by where the fix lives:

**This PR (CLI)**

- **`search hybrid` never actually runs hybrid.** It sends the file to `/search` with no `mode`, but the server only goes hybrid when the body has `mode:"hybrid"`. So a `{"queries":[...]}` file just errors with `exactly one of 'query','vector','sparse_vector'`, and if you add a top-level `query` to get past that it quietly runs plain dense — which is why the full-text lane looked like it did nothing. The only way it works today is hand-adding `mode:"hybrid"`, and nothing tells you to. Injecting `mode` in the command + documenting the lane shape would fix it (the full-text lane is `full_text:"<text>"`, not `using:"full_text"`).
- **`chunks list` has no `--filter`,** even though `count` and `bulk-delete` do — and the server's list handler already reads a `filter` query param. Just needs wiring into `ListChunksOpts`.
- **None of the `--file` bodies are documented.** Six commands take `--file` and I reverse-engineered every shape from validation errors. One example in each command's help would cover it. (`create --help` also never mentions the immutable fields — only `patch` does.)

**Operator repo, not this PR — noting for the record**

- `document_id` is silently dropped on upsert. The field is `documentId`, so the snake_case guess is ignored and every chunk ends up as its own document — reported as success.
- Bare-list filter `{"cat":["a","b"]}` returns a 500 instead of a "use `$in`" message.
- Creating against a still-provisioning DB returns a generic 500.
- `documents list` only returns id + chunk count — no metadata, no filter.

dgallego-iai:
Thanks for the detailed run-through — all three CLI items are addressed on the branch:

- **`search hybrid` never going hybrid**: the command now injects `mode:"hybrid"` into the body itself (37e1269), so a plain `{"queries":[...]}` file works as-is — no hand-editing. The help documents the lane shape (`full_text:"<text>"`, not `using:"full_text"`).
- **`chunks list --filter`**: wired through `ListChunksOpts` to the server's `filter` query param (e79feb3).
- **Undocumented `--file` bodies**: added an `iai collections schema` command covering the collection config, chunk records, and search/hybrid file bodies (0c7eefa); every `--file` flag's help now points at it, and `create --help` lists the immutable fields (54c59ce).

The operator-side items you noted for the record are all fixed in Interactive-AI-Labs/deployment-operator#206: unknown/misspelled fields (e.g. `document_id`) are now a 400 that names the field instead of a silent drop, bare-list filters get a 400 suggesting `$in`, creating against a provisioning DB returns a 503 "not ready" instead of a generic 500, and `documents list` takes a metadata `--filter`.

internal/clients/collections.go
```diff
@@ -0,0 +1,260 @@
...
+
+// CollectionSlot describes one vector slot in a collection's config.
+type CollectionSlot struct {
+	Type      string `json:"type"`
+	Dimension int    `json:"dimension"`
+	Distance  string `json:"distance"`
+	Embedding *struct {
+		Model string `json:"model"`
+	} `json:"embedding"`
+	Index *struct {
```
interactivealex: This only decodes `type`, so `describe` (table, `--json`, and `--yaml`) drops `ef_search_default`/`m`/`ef_construct`. Since `patch` and `slots add` can set those, there's no way to confirm a patch actually landed. Could we keep the rest of the index config here (extra fields or a `map[string]any`)?

cmd/collections_documents.go
```diff
@@ -0,0 +1,158 @@
...
+		_ = c.MarkFlagRequired("database")
+	}
+
+	for _, c := range []*cobra.Command{documentsListCmd, documentsGetCmd} {
+		c.Flags().BoolVar(&collJSON, "json", false, "Output raw API response as JSON")
+		c.Flags().BoolVar(&collYAML, "yaml", false, "Output raw API response as YAML")
+		c.Flags().IntVar(&docLimit, "limit", 0, "Page size (1-1000, default 100)")
+		c.Flags().StringVar(&docCursor, "cursor", "", "Opaque cursor from a previous page")
+		c.Flags().
+			StringVar(&docFilter, "filter", "", "Metadata filter as a JSON object (list only)")
```
interactivealex: `--filter` gets registered on `get` too, but `GetDocument` never reads it (the help even says "list only"). So `documents get ... --filter` is silently ignored. Let's only register it on the list command.

internal/inputs/collections.go
```diff
@@ -0,0 +1,97 @@
...
+// JSON, so both parse; yaml.v3 decodes mappings into string-keyed maps, so the
+// result re-marshals cleanly to JSON.
+func ReadCollectionBodyFile(path string) ([]byte, error) {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read %q: %w", path, err)
+	}
+
+	var doc any
+	if err := yaml.Unmarshal(data, &doc); err != nil {
```
interactivealex, line 20: Two edge cases here: a numeric YAML key (e.g. `2024:` in metadata) decodes to `map[interface{}]interface{}` and the `json.Marshal` below then fails with a cryptic `json: unsupported type` — the comment says that can't happen but it can. Also, an empty or comment-only file marshals to `null` and we send that as the body. Worth rejecting empty input and giving a clearer message on non-string keys.

internal/output/collections_chunks.go
```diff
@@ -0,0 +1,102 @@
...
+	fmt.Fprintf(out, "ID:        %s\n", c.ID)
+	fmt.Fprintf(out, "Document:  %s\n", c.DocumentID)
+	fmt.Fprintf(out, "Text:      %s\n", c.Text)
+
+	if len(c.Metadata) > 0 {
+		b, _ := json.MarshalIndent(c.Metadata, "           ", "  ")
+		fmt.Fprintf(out, "Metadata:  %s\n", string(b))
+	}
+
+	for slot, raw := range c.Vectors {
```
interactivealex: Two small things: this ranges the map unsorted so multi-slot output order is random (the other printers sort their keys), and the `err == nil` on the next line silently drops any vector that isn't a plain float array (sparse/binary slots), so it looks absent. Maybe sort the slots and print a note instead of dropping.

cmd/collections.go
```diff
@@ -0,0 +1,409 @@
...
+		}
+		if collYAML {
+			return output.PrintStructuredYAML(out, collectionSchemas())
+		}
+		fmt.Fprint(out, collectionSchemaText)
+		return nil
+	},
+}
+
+const collectionSchemaText = `Collection file schemas (bodies for --file). All fields camelCase.
```
interactivealex: This says "All fields camelCase" but the examples right below (and the server) use snake_case: `full_text`, `ef_search_default`, `candidate_limit`. Anyone who trusts the banner will camelCase these and they'll be silently ignored. Let's drop or fix that line.

cmd/collections_documents.go
```diff
@@ -0,0 +1,158 @@
...
+	docSubcommands := []*cobra.Command{documentsListCmd, documentsGetCmd, documentsDeleteCmd}
+	for _, c := range docSubcommands {
+		c.Flags().StringVarP(&collOrganization, "organization", "o", "", "Organization name")
+		c.Flags().StringVarP(&collProject, "project", "p", "", "Project name")
+		c.Flags().
+			StringVarP(&collDatabase, "database", "d", "", "Database that holds the collection (required)")
+		_ = c.MarkFlagRequired("database")
+	}
+
+	for _, c := range []*cobra.Command{documentsListCmd, documentsGetCmd} {
```
interactivealex: `documents delete` isn't in this loop, so it has no `--json`/`--yaml` — even though it returns a structured `{documentId, deletedCount}`. `chunks bulk-delete` (same kind of result) does support them. Worth adding here for scripting.

internal/output/collections.go
```diff
@@ -0,0 +1,112 @@
...
+	const unit = 1024
+	if n < unit {
+		return fmt.Sprintf("%d B", n)
+	}
+	div, exp := int64(unit), 0
+	for v := n / unit; v >= unit; v /= unit {
+		div *= unit
+		exp++
+	}
+	return fmt.Sprintf("%.1f %ciB", float64(n)/float64(div), "KMGTPE"[exp])
```
interactivealex: The unit is picked from the integer quotient before `%.1f` rounds, so `1048575` prints as "1024.0 KiB" instead of "1.0 MiB" (same at each 1024^k boundary). Easy to fix by re-checking the value after dividing.

cmd/collections_slots.go
```diff
@@ -0,0 +1,277 @@
...
+	slotsAddCmd.Flags().StringVar(
+		&slotType,
+		"type",
+		"float32",
+		"Vector slot type (float32, float16, binary, or sparse; default: float32)",
+	)
+	slotsAddCmd.Flags().
+		IntVar(&slotDimension, "dimension", 0, "Vector dimension (required unless --file is provided)")
+	slotsAddCmd.Flags().
+		StringVar(&slotDistance, "distance", "", "Distance metric (default: cosine)")
```
interactivealex: Help says "(default: cosine)" but the flag default is empty and we don't fill in cosine — the server does. Same with `--using` ("default: default"). Either set the default here or reword the help so it's accurate.

cmd/collections_search.go
```diff
@@ -0,0 +1,260 @@
...
+	searchByIDCmd.Flags().IntVar(&searchLimit, "limit", 0, "Max results")
+	searchByIDCmd.Flags().
+		BoolVar(&searchExcludeSelf, "exclude-self", false, "Exclude the seed chunk")
+	searchByIDCmd.Flags().StringVar(&searchFilter, "filter", "", "Metadata filter as a JSON object")
+
+	searchHybridCmd.Flags().
+		StringVar(&searchFile, "file", "", "Path to a YAML/JSON hybrid-search file")
+	_ = searchHybridCmd.MarkFlagRequired("file")
+
+	searchCmd.AddCommand(searchBatchCmd, searchByIDCmd, searchHybridCmd)
```
interactivealex, line 262: Because `search` is runnable and also parents `batch`/`by-id`/`hybrid`, a collection named `batch`/`by-id`/`hybrid` can't be searched via `search` — cobra routes to the subcommand instead. Rare, but a note or guard would save a confusing "unknown flag" error.

internal/clients/collections.go
```diff
@@ -0,0 +1,260 @@
...
+		return "", fmt.Errorf("failed to create request: %w", err)
+	}
+
+	resp, err := c.do(req)
+	if err != nil {
+		return "", fmt.Errorf("delete collection request failed: %w", err)
+	}
+	defer resp.Body.Close()
+
+	respBody, _ := io.ReadAll(resp.Body)
```
interactivealex: This (and DeleteChunk/DeleteSlot/sendCollectionBody) does `respBody, _ :=` and drops the read error, but `collectionErr` deliberately folds it in. A truncated body on a network blip will read as a clean rejection here. Worth staying consistent with `collectionErr`.

internal/output/collections_chunks.go
```diff
@@ -0,0 +1,102 @@
...
+	if len(r.DeletedIds) > 0 {
+		for _, id := range r.DeletedIds {
+			fmt.Fprintf(out, "  %s\n", id)
+		}
+	}
+	return nil
+}
+
+// truncate shortens s to n runes, appending an ellipsis when cut.
+func truncate(s string, n int) string {
```
interactivealex: We already have `truncateString` (comments.go) and `truncateValue` (summary.go) in this package — this is a third truncation helper with slightly different behavior, so collection tables truncate differently from the rest of the CLI. Let's reuse one of the existing ones.

internal/clients/collections_slots.go
```diff
@@ -0,0 +1,179 @@
...
+
+func slotPath(orgId, projectId, database, collection, slot string) string {
+	base := collectionsPath(orgId, projectId, database)
+	return base + "/" + url.PathEscape(collection) + "/vectors/" + url.PathEscape(slot)
+}
+
+// sendSlotBody issues a POST/PUT request to a slot path with an optional JSON
+// body and decodes the response into dst. Use an inlined GET for read-only
+// slot endpoints to keep the body-bearing path explicit.
+func (c *DeploymentClient) sendSlotBody(
```
interactivealex: `sendSlotBody` is almost identical to `postJSONInto` in collections_search.go. Could we parameterize that one with a method + nil-body guard and drop this copy?

internal/inputs/collections.go
```diff
@@ -0,0 +1,97 @@
...
+	return body, nil
+}
+
+// BuildChunkCountBody builds the count request body from an optional metadata
+// filter (a JSON object string) and an optional id prefix.
+func BuildChunkCountBody(filterJSON, prefix string) ([]byte, error) {
+	body := map[string]any{}
+	if filterJSON != "" {
+		var filter map[string]any
+		if err := json.Unmarshal([]byte(filterJSON), &filter); err != nil {
```
interactivealex: This filter-parse block is the same as `addFilter` in collections_search.go (and it's inlined again in BuildBulkDeleteBody). Worth calling `addFilter` so the validation/message stays in one place.

internal/clients/collections.go
dgallego-iai: Good catch — the client now decodes the full index object (`m`, `ef_construction`, `lists`, `ef_search_default`), so `--json`/`--yaml` carry everything and the table shows the parameters inline, e.g. `hnsw (m=16, ef_construction=64, ef_search_default=80)`. A patched `ef_search_default` is now visible in describe. 47e02de

cmd/collections_documents.go
dgallego-iai: Done — `--filter` is now registered only on `documents list`, so `documents get ... --filter` fails with "unknown flag" instead of being silently ignored. Docs regenerated. 16ef47b

internal/inputs/collections.go
dgallego-iai: Both confirmed with a quick probe (yaml.v3 really does fall back to `map[interface{}]interface{}` on a non-string key). Fixed in 74aeaf3:

- empty / comment-only file → rejected with `"…" has no content; see 'iai collections schema'…` instead of sending a `null` body.
- non-string keys → stringified recursively (`2024:` becomes `"2024"`) rather than erroring, since JSON keys are always strings anyway — that is what the author meant.

Table test added covering both plus the happy path.

internal/output/collections_chunks.go
dgallego-iai: Fixed in 6152679 — slots print sorted by name, and instead of dropping non-dense vectors each kind gets a one-line summary: `2 dims` (dense), `sparse, 2 of 100 dims set`, `binary, 4 bits`, with a `present (unrecognized format)` fallback so nothing ever disappears silently. Test asserts the sorted multi-slot output.

cmd/collections.go
dgallego-iai: Fixed the banner rather than dropping it — it now states the real rule: keys are snake_case (`full_text`, `ef_search_default`) except chunk record fields, which are camelCase (`documentId`). acd637f

cmd/collections_documents.go
dgallego-iai: Added — `documents delete` now takes `--json`/`--yaml` and emits the raw `{documentId, deletedCount}`, same pattern as `chunks bulk-delete`. 2c64082

docs/iai_collections.md
```diff
@@ -0,0 +1,44 @@
+## iai collections
+
+Vector collections (knowledge bases) inside a pgvector database
+
+### Synopsis
+
+Manage vector collections within a database.
+
+A collection is a vector store (knowledge base) that lives inside an existing
```
ajaen-ai: "A collection is a vector store (knowledge base)" -> to be for an external user this won't be clear right away?

Can we say, using the correct wording, something like "table of chunks"? Not sure here about the correct way but that phrase wasn't intuitive for me

internal/output/collections.go
dgallego-iai: Nice catch. Fixed as you suggested — after picking the unit, if `%.1f` would render 1024.0 (value ≥ 1023.95) it steps up one unit, so 1048575 → `1.0 MiB` at every 1024^k boundary. Boundary cases added to the table test, including 1048524 → `1023.9 KiB` staying put. 0344d0f

docs/iai_collections.md
ajaen-ai: In docs/iai_collections.md:

> Manage the chunks (rows) in a collection

cmd/collections_slots.go
dgallego-iai: Went with the reword: the help now says "omit for the server default, cosine" (and the same for `--using`). Kept the server as the owner of the default rather than baking it into the flag, so an older CLI can never pin a stale default if the server's ever changes. 5508d47

docs/iai_collections_slots.md
```diff
@@ -0,0 +1,32 @@
+## iai collections slots
+
+Manage a collection's vector slots and their indexes
```
ajaen-ai, line 3: I think we need to also describe what a slot is, adding what the comparison is to a chunk

internal/clients/collections.go
dgallego-iai: Fixed for all four (`DeleteCollection`, `DeleteChunk`, `DeleteSlot`, `sendCollectionBody`) in 46188d7. They now route through a shared `serverMessage` helper: error path reuses `collectionErr` (which already folds the read error in), success path surfaces the `io.ReadAll` error instead of dropping it. That also collapses the four duplicated status-check blocks into one.

interactivealex (approved):
All correctness, consistency, and doc findings addressed and verified — index config now decodes fully, YAML empty/non-string keys handled, humanBytes boundary fixed, dead `--filter` removed, schema banner corrected. Build + tests green. The remaining reuse/dedup notes (truncate, sendSlotBody/postJSONInto, addFilter, delete-path read errors) are non-blocking nits — fine to defer. LGTM 🚀

internal/clients/collections.go
```diff
@@ -0,0 +1,268 @@
...
+
+// PatchCollection updates a collection's mutable config from a raw JSON body.
+func (c *DeploymentClient) PatchCollection(
+	ctx context.Context,
+	orgId, projectId, database, name string,
+	body []byte,
+) (string, error) {
+	path := collectionsPath(orgId, projectId, database) + "/" + url.PathEscape(name)
+	return c.sendCollectionBody(ctx, http.MethodPatch, path, body, "update collection", "")
+}
```
ajaen-ai: This func is only used onced and its two lines, I think it doesn't add value.

I would inline it

internal/output/collections_chunks.go
dgallego-iai: Done in fcdd261 — dropped the local `truncate` and pointed the chunk/search/document table cells at `truncateString` (the one the comments-list table already uses), so collection tables truncate like the rest of the CLI. Went with `truncateString` over `truncateValue` since the latter appends " (truncated)", too verbose for a table cell. Repurposed the old test to cover `truncateString`, which previously had none.

internal/clients/collections_search_test.go
ajaen-ai, line 1: Can we do this table based?

internal/inputs/collections_test.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func TestBuildAddSlotBody(t *testing.T) {
+	if _, err := BuildAddSlotBody("float32", 1536, "cosine"); err != nil {
+		t.Errorf("valid slot body should not error: %v", err)
+	}
+	if _, err := BuildAddSlotBody("float32", 0, ""); err == nil {
+		t.Errorf("dimension 0 should error")
+	}
+}
```
ajaen-ai, line 172: I think we can table base this as well

internal/output/collections_documents.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	if doc.HasMore && doc.NextCursor != nil {
+		fmt.Fprintf(out, "\nMore chunks — next page: --cursor %s\n", *doc.NextCursor)
+	}
+	return nil
+}
+
+// PrintDeleteDocumentResult renders a document-delete response.
+func PrintDeleteDocumentResult(out io.Writer, r *clients.DeleteDocumentResult) error {
```
ajaen-ai: I think we don't need this func

internal/output/collections.go
```diff
@@ -0,0 +1,135 @@
...
+	for i, c := range collections {
+		rows[i] = []string{c.Name, LocalTime(c.CreatedAt), LocalTime(c.UpdatedAt)}
+	}
+	return PrintTable(out, headers, rows)
+}
+
+// PrintCollectionDescribe renders a collection's config: header fields plus a
+// per-slot table.
+func PrintCollectionDescribe(out io.Writer, c *clients.DescribeCollectionResponse) error {
+	fmt.Fprintf(out, "Name:       %s\n", c.Name)
```
ajaen-ai: This is describe-style output, so we should use `NewDescribeWriter(out)` and `\t` between label/value instead of manually padded spaces. That keeps indentation consistent with services/traces/etc.

internal/output/collections.go
```diff
@@ -0,0 +1,135 @@
...
+			embeddingModel(slot),
+			indexType(slot),
+		})
+	}
+	return PrintTable(out, headers, rows)
+}
+
+// PrintCollectionStats renders a collection's operational stats.
+func PrintCollectionStats(out io.Writer, s *clients.CollectionStats) error {
+	fmt.Fprintf(out, "Chunks:  %d\n", s.ChunkCount)
```
ajaen-ai: Same convention issue here: prefer `NewDescribeWriter` + tab-separated labels over hardcoded spacing.

internal/output/collections_chunks.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// PrintChunk renders a single chunk's detail.
+func PrintChunk(out io.Writer, c *clients.Chunk) error {
+	fmt.Fprintf(out, "ID:        %s\n", c.ID)
+	fmt.Fprintf(out, "Document:  %s\n", c.DocumentID)
+	fmt.Fprintf(out, "Text:      %s\n", c.Text)
+
+	if len(c.Metadata) > 0 {
+		b, _ := json.MarshalIndent(c.Metadata, "           ", "  ")
```
ajaen-ai: Avoid custom `json.MarshalIndent(..., "           ", "  ")`. Existing convention is pretty JSON plus `indentLines(...)`, so multiline JSON indentation stays standardized.

internal/output/collections_slots.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"fmt"
+	"io"
+
+	"github.com/Interactive-AI-Labs/interactive-cli/internal/clients"
+)
+
+// PrintSlotAddResult renders the result of adding a slot.
+func PrintSlotAddResult(out io.Writer, r *clients.SlotAddResult) error {
+	fmt.Fprintf(out, "Slot:          %s\n", r.Slot)
```
ajaen-ai: These result/detail outputs use manual alignment. Please switch to `NewDescribeWriter` with tabs so spacing is generated consistently.

internal/output/collections_documents.go
```diff
@@ -0,0 +1,59 @@
...
+
+	if list.HasMore && list.NextCursor != nil {
+		fmt.Fprintf(out, "\nMore results — next page: --cursor %s\n", *list.NextCursor)
+	}
+	return nil
+}
+
+// PrintDocumentChunks renders one document's chunks: a header plus a chunk table.
+func PrintDocumentChunks(out io.Writer, doc *clients.DocumentChunks) error {
+	fmt.Fprintf(out, "Document:  %s\n\n", doc.DocumentID)
```
ajaen-ai: This header is manually padded; use the describe writer convention here too.

internal/output/collections_test.go
```diff
@@ -0,0 +1,569 @@
...
+
+func TestPrintSlotResults(t *testing.T) {
+	t.Run("add", func(t *testing.T) {
+		var buf bytes.Buffer
+		if err := PrintSlotAddResult(&buf, &clients.SlotAddResult{
+			Slot: "s", Type: "float32", Dimension: 4, Distance: "l2", IndexStatus: "ready",
+		}); err != nil {
+			t.Fatal(err)
+		}
+		want := "Slot:          s\nType:          float32\nDimension:     4\nDistance:      l2\nIndex status:  ready\n"
```
ajaen-ai: These expected strings lock in manual spacing. Once output uses `NewDescribeWriter`, update expectations to the standardized tabwriter output.

internal/clients/collections_slots.go
dgallego-iai: Merged in ad99752 — both are now one `sendJSONInto(ctx, method, path, body, action, dst)` in collections.go (the shared home next to `collectionErr`/`serverMessage`), with the nil-body guard so the bodyless vacuum POST still works. `postJSONInto` and `sendSlotBody` are gone; search and slot callers pass their own method.

internal/inputs/collections.go
dgallego-iai: Done in b17ab79 — both `BuildChunkCountBody` and `BuildBulkDeleteBody` now call `addFilter`, so the parse and the "--filter must be a JSON object" message live in one place.

cmd/collections_search.go
dgallego-iai: Went with a note rather than a guard (dc5f595) — `search --help` now warns that a collection named batch/by-id/hybrid is shadowed by the sub-command and to rename it or use the API. A real guard would mean intercepting cobra's sub-command resolution, which fights the framework for a very rare case; the note keeps behaviour predictable and explains the "unknown flag" if anyone hits it.

docs/iai_collections.md
dgallego-iai: Agreed — reworded (495efad). It now leads with the concrete framing: "A collection is a table of chunks (rows) — each chunk is text plus its vector embedding(s) — that you search by meaning or keyword; it's what backs a knowledge base." Dropped "vector store" from the short desc too.

dgallego-iai: Adopted the same "(rows)" style you flagged on the chunks command — the collection description now uses "table of chunks (rows)". 495efad

docs/iai_collections_slots.md
dgallego-iai: Added a definition (59cc63c): "A slot is a named vector space (a column) on a collection: if a collection is a table and a chunk is a row, a slot is a vector column down every row. A collection can have several — e.g. a dense slot for embeddings and a sparse slot for keywords — and each chunk holds one vector per slot."

internal/clients/collections.go
dgallego-iai: Inlined (1ca1ecd) — `PatchCollection` is gone; the patch command builds the path and calls `SendCollectionBody` directly. Exported `CollectionsPath` + `SendCollectionBody` (create still uses both) so the URL template stays in one place rather than being duplicated in cmd.

internal/clients/collections_search_test.go
dgallego-iai: Done (543d2e7) — `TestWithHybridMode` is now table-driven (name/body/want/wantErr) with `cmp.Diff` on the decoded body, covering the three cases (inject when absent, force over an existing mode, non-object error).

internal/inputs/collections_test.go
dgallego-iai: Done (next commit) — `TestBuildChunkCountBody` and `TestBuildAddSlotBody` are now name/want/wantErr tables, and they assert the actual produced body (e.g. `{"filter":{"lang":"en"},"prefix":"doc-"}`) rather than just error-vs-no-error. The rest of the file (ParseVector, BuildBulkDeleteBody, BuildSearchBody, ReadCollectionBodyFile) was already table-based.

internal/output/collections_documents.go
dgallego-iai: Agreed — it was a one-line fmt.Fprintf used once. Inlined into the documents delete command and dropped the helper (next commit). Kept PrintDocumentList/PrintDocumentChunks since those build tables and pagination.

internal/output/collections.go
dgallego-iai: Done (d643166) — `PrintCollectionDescribe` now uses `NewDescribeWriter` with tab-separated labels; the header aligns via the tabwriter like databases/services describe.

internal/output/collections.go
dgallego-iai: Done (d643166) — `PrintCollectionStats` switched to `NewDescribeWriter` + tabs as well.

internal/output/collections_chunks.go
dgallego-iai: Done (d643166) — dropped the custom `MarshalIndent(..., "           ", "  ")`; chunk metadata now prints via standard pretty JSON + `indentLines("  ")`, so multiline indentation is standardized.

internal/output/collections_slots.go
dgallego-iai: Done (d643166) — `PrintSlotAddResult` and `PrintSlotIndexProgress` now use `NewDescribeWriter` + tabs. (`PrintSlotOpResult` is a single "Slot %q: %s" line, no alignment needed, so left as-is.)

internal/output/collections_documents.go
dgallego-iai: Done (d643166) — the `PrintDocumentChunks` header now goes through `NewDescribeWriter` instead of a hand-padded "Document:  %s".

internal/output/collections_test.go
dgallego-iai: Updated (d643166) — the expectations no longer lock manual spacing: added a `squashSpaces` helper so the tests assert content with whitespace normalized, immune to whatever column widths the tabwriter picks.

ajaen-ai:
Pre-commit is failing, remember to check all pipelines go through before asking for rereview 😉

ajaen-ai:
I would squash commits (52 in this PR) to whatever number/segregation makes sense

ajaen-ai (approved):

interactivealex:
Looks good to me

PR #156, opened 2026-06-30
fix(schema): name discriminated unions in agents schema TYPE column

interactivealex:
## Problem

`iai agents schema` renders the `search` row's TYPE column as `any | null` instead of `KnowledgeBase | ExternalSearch | null`. Display-only bug — the fetched JSON schema is correct; the Go renderer just can't name the type.

## Root cause

The type-namer (`resolveTypeName` in `internal/output/schema.go`) walked `anyOf` members but had **no `oneOf` case**. Pydantic emits a discriminated union as a `oneOf` nested inside an `anyOf`:

```json
"anyOf": [
  { "oneOf": [ {"$ref": "KnowledgeBase"}, {"$ref": "ExternalSearch"} ],
    "discriminator": { "propertyName": "type", "mapping": {...} } },
  { "type": "null" }
]
```

The namer stripped the null branch, then hit the `oneOf` wrapper — no `type`, no top-level `$ref` — and fell through to its `any` default.

## Fix

Add `OneOf` to the schema struct and resolve `oneOf` members the same way as `anyOf`: recurse each to its `$ref` title and join with `" | "`. General fix — corrects **every** discriminated union, not just `search`.

Same class of bug as the recently-fixed `gen_docs.py` issue, but in the CLI's Go renderer.

## Test

Extended `TestPrintSchemaPretty` with a `search`-shaped discriminated-union field; asserts `KnowledgeBase | ExternalSearch | null`.

### DIFF
--- a/internal/output/schema.go
+++ b/internal/output/schema.go
@@ -18,6 +18,7 @@ type jsonSchema struct {
 	Ref         string                `json:"$ref"`
 	Items       *jsonSchema           `json:"items"`
 	AnyOf       []jsonSchema          `json:"anyOf"`
+	OneOf       []jsonSchema          `json:"oneOf"`
 	Enum        []any                 `json:"enum"`
 	Default     any                   `json:"default"`
 	Const       any                   `json:"const"`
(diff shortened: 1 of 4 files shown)


interactivealex:
Trimmed the comment to one line and dropped the `ponytail:` tag (my mistake). Bumped to 0.34.7. @ajaen-ai ready when you are.

internal/output/schema.go
```diff
@@ -113,9 +114,13 @@ func resolveTypeName(s jsonSchema, defs map[string]jsonSchema) string {
 		return refName(s.Ref)
 	}
 
-	if len(s.AnyOf) > 0 {
-		parts := make([]string, 0, len(s.AnyOf))
-		for _, alt := range s.AnyOf {
+	// anyOf and oneOf are unioned the same way; discriminated unions nest a
+	// oneOf inside an anyOf, so recursion resolves each member to its $ref name.
+	// ponytail: Pydantic never puts both on one object, so we flatten them; if a
+	// schema ever did, the members merge without grouping (A | B | C | D).
```
ajaen-ai: Remember to keep comments concise.

Long comments are ignored and usually have more information than needed.

They wont be maintained and will create noise early.

interactivealex: fixed

ajaen-ai:
Remember to only pass PR to be reviewed when completely ready, this one has conflicts

internal/output/schema.go
ajaen-ai: (Non blocking)

ajaen-ai (approved):

PR #160, opened 2026-07-09
feat: iai mcps command tree + --mcp on agents (DEV-840)

dgallego-iai:
## What

The CLI surface for [DEV-840](https://linear.app/interactive-ai/issue/DEV-840/integrate-connectors-and-agents-better): `iai mcps` — thin wrappers over the deployment-operator's `/mcps` endpoints (Interactive-AI-Labs/deployment-operator#215), mirroring the `iai services`/`iai agents` command patterns.

- **`iai mcps`**: `catalog`, `create`, `update`, `list`, `get`, `verify` (external only), `run-tool`, `delete` (blocks on attached agents; `-f` to force), and **`tools get [--revision N]` / `tools revisions` / `tools diff <a> <b>`** — the tools cache is versioned per helm revision on the operator side, and the revisions/diff commands reuse the existing `agents revisions`/`agents diff` machinery (`ParseRevisionArg`, `PrintRevisionDiff`).
- **Agents integration**: `--mcp <name>` (repeatable) on `iai agents create`/`update` injects a bare `{mcp_id: name}` ref (resolved by the operator at deploy time); `--detach-mcp <name>` removes one. On `update` without `--file`, the current config is fetched via describe and the mcp change is overlaid onto the same patch as every other flag — combining `--mcp` with `--env`/`--secret`/etc. works.
- **BREAKING: the old `iai connectors` command tree is removed** (platform-backed `mcp_connection` CRUD/verify/run-tool), superseded by `iai mcps`. CLI-only removal — the platform's `/mcp-connections` endpoints stay for the UI. The platform client keeps only `ListMcpCatalog`, which `mcps catalog` uses.

## Validated

- Full test battery against the local binary + local operator + dev platform: catalog browse, internal/external/catalog-backed create (including a real GitHub PAT dial), list/get/verify/run-tool, tools revisions + diff across an update, no-`--file` attach/detach, credential rotation restarting the mcp and every attached agent.
- Operator-side e2e (`e2e/mcps.sh`) 16/16 with this branch's flows.
- Post-connectors-removal smoke test: `iai connectors` → unknown command; `mcps catalog`/`mcps list` green against dev.

## Local run

```sh
go build -o /tmp/iai-local ./cmd/iai
alias iai-local="/tmp/iai-local --hostname https://dev.interactive.ai --deployment-hostname http://localhost:8000"
```


dgallego-iai:
Went through the review points:

- **Bug: `--mcp` without `--file` silently drops other flags** — fixed in 30a0961. The mcp-derived patch is now overlaid onto the same patch `BuildAgentUpdatePatch` builds for every other flag, instead of replacing it.
- **Missing `MarkFlagsMutuallyExclusive(\"json\", \"yaml\")`** — fixed in 78fb7f1, across all 5 commands that have the pair (the 4 named here plus `mcp tools get`, added after this review ran).
- **`PutMcp` silently renders empty strings in the restarted list** — fixed in 060d809, non-string entries are now skipped instead of becoming blank.
- **`orDash` is package-private to `mcps.go`** — turned out to be a non-issue: Go privacy is per-package, not per-file, so it was already callable from any file in `internal/output`. Since it only had one call site anyway, inlined it instead in 513ece6.
- **Redundant mutual-exclusivity check for `--catalog-id`/`--endpoint-url`** — confirmed dead (cobra's `MarkFlagsMutuallyExclusive` already rejects it first) and deleted in f5f4542.
- **No way to detach an MCP without `--file`** — already addressed earlier in the branch's history (`--detach-mcp`), landed after this review ran.

cmd/agents.go
```diff
@@ -888,6 +925,8 @@ func init() {
 		StringVar(&agentScheduleTimezone, "schedule-timezone", "", "IANA timezone for the schedule (e.g. Europe/Berlin, US/Eastern, UTC); required with --schedule-uptime or --schedule-downtime")
 	agentCreateCmd.Flags().
 		StringVar(&agentStackId, "stack-id", "", "Stack ID to assign the agent to")
+	agentCreateCmd.Flags().
+		StringArrayVar(&agentMcpNames, "mcp", nil, "Attach an MCP by name (see 'iai mcps list'); can be repeated")
```
ajaen-ai, line 929: Since this is a list, should it be plural (mcps)

cmd/agents.go
```diff
@@ -925,6 +964,10 @@ func init() {
 		StringVar(&agentStackId, "stack-id", "", "Stack ID to assign the agent to")
 	agentUpdateCmd.Flags().
 		BoolVar(&agentClearStackId, "clear-stack-id", false, "Remove the agent from its stack")
+	agentUpdateCmd.Flags().
+		StringArrayVar(&agentMcpNames, "mcp", nil, "Attach an MCP by name (see 'iai mcps list'); can be repeated. Without --file, appends to the agent's current mcps")
+	agentUpdateCmd.Flags().
+		StringArrayVar(&agentDetachMcpNames, "detach-mcp", nil, "Detach an MCP by name; can be repeated. Without --file, removes from the agent's current mcps (applied before --mcp)")
```
ajaen-ai, line 970: Same here

docs/iai_mcps_create.md
```diff
@@ -0,0 +1,79 @@
+## iai mcps create
+
+Create an mcp in a project
+
+### Synopsis
+
+Create an mcp — an in-cluster MCP server ("internal"), a custom external
+URL, or a catalog-backed provider.
+
+Internal: --image-name, --image-tag, --port (like 'iai services create'); --env
```
ajaen-ai: I wouldn't add references to other non related commands like iai services create.

It doesn't add much value: it can confuse the agent to read other commands it doesn't need and adds more text to the docs.

docs/iai_mcps_create.md
```diff
@@ -0,0 +1,79 @@
...
+and --secret load env vars from literal values or existing k8s Secrets. --path
```
ajaen-ai: Careful with exposing internal stack details, k8s in this case

docs/iai_mcps_create.md
```diff
@@ -0,0 +1,79 @@
...
+is the endpoint path the mcp's own server exposes (default "/mcp" — a common
+convention, not something the MCP protocol requires, so set it to whatever the
+mcp owner actually configured rather than assuming).
```
ajaen-ai: I think we can say this much more concise

docs/iai_mcps_create.md
```diff
@@ -0,0 +1,79 @@
...
+External custom: --external-url — a server not owned by the platform, dialed
+directly at that URL, path included.
+External catalog: --catalog-id (see 'iai mcps catalog'); external URL and auth are
+derived from the catalog entry. Pass an auth type the entry supports; catalog
+entries provide their own credential header and prefix.
+
+The mcp is verified against the live server before it's kept: an internal mcp
+is verified once its pod is up (checked in the background — see 'iai mcps get');
```
ajaen-ai: Same here, I think we can reword this whole text better (and careful we use replica instead of pod)

"An internal mcp is verified once its status is healthy. An extenral mcp is verified immediately. The create fails if the mcp is unreachable or refects credentials"

docs/iai_mcps_get.md
```diff
@@ -1,21 +1,21 @@
-## iai connectors get
+## iai mcps get
```
ajaen-ai: We use describe for live components (agents, databases, ...)

We previously used get because this was just data

docs/iai_mcps_tools.md
```diff
@@ -0,0 +1,43 @@
+## iai mcps tools
+
+List an mcp's cached tools with descriptions
+
+### Synopsis
+
+Show the full cached tool list — name and description. 'iai mcps get' only
```
ajaen-ai: Don't we show the tool arguments as well?

docs/iai_mcps_update.md
```diff
@@ -0,0 +1,64 @@
+## iai mcps update
+
+Replace an mcp's spec
+
+### Synopsis
+
+Full replace — pass the mcp's complete desired spec, same flags as create.
+There is no partial-update mechanism for mcps; anything not passed resets to its
```
ajaen-ai: There is no partial update mechanism? Don't we have a patch endpoint like we do for agents and services?

internal/output/mcps_test.go
```diff
@@ -0,0 +1,28 @@
...
+			AuthMethods: []string{"api_key"},
+		},
+	}
+	if err := PrintMcpCatalog(&buf, entries); err != nil {
+		t.Fatalf("error = %v", err)
+	}
+	if !strings.Contains(buf.String(), "GitHub") || !strings.Contains(buf.String(), "e1") {
+		t.Fatalf("unexpected output:\n%s", buf.String())
+	}
+}
```
ajaen-ai, line 43: Can we follow the pattern we use in other output tests in this repo?

expected == real output

internal/inputs/mcp_create_test.go
```diff
@@ -0,0 +1,178 @@
...
+	t.Run("internal with no --path leaves it empty for the server to default", func(t *testing.T) {
+		body, err := BuildMcpRequestBody(McpInput{Port: 8080, ImageName: "my-mcp", ImageTag: "v1"})
+		if err != nil {
+			t.Fatalf("unexpected err: %v", err)
+		}
+		if body.Path != "" {
+			t.Errorf("Path = %q, want empty (CLI must not guess)", body.Path)
+		}
+	})
+}
```
ajaen-ai: Lets follow the patterns from the rest of the repo here

internal/clients/deployment_client_mcps.go
```diff
@@ -0,0 +1,398 @@
...
+	ToolsAdded          []string `json:"toolsAdded,omitempty"`
+	ToolsRemoved        []string `json:"toolsRemoved,omitempty"`
+	ChangedFromRevision string   `json:"changedFromRevision,omitempty"`
+}
+
+type listMcpsResponse struct {
+	Mcps []McpOutput `json:"mcps"`
+}
+
+type RunMcpToolResult struct {
```
ajaen-ai, line 105: Nit: rather than adding an MCP-specific `sendMcpRequest`, can we make this a generic deployment client helper?

Something like `sendJSONRequest(ctx, method, path, body)` could:

- create the request
- optionally marshal/set JSON body
- send it
- read the response body
- surface server `message` on non-2xx

Then MCP methods can call that directly, and we avoid a resource-specific helper that mostly duplicates logic we could reuse for agents/services/databases later.

internal/inputs/mcp_create_test.go
```diff
@@ -0,0 +1,178 @@
...
+import (
+	"strings"
+	"testing"
+)
+
+func TestBuildMcpRequestBody(t *testing.T) {
+	tests := []struct {
+		name       string
+		input      McpInput
+		wantErr    bool
```
ajaen-ai: Nit: this can be flatter if `wantErr` is the exact expected error string instead of separate `wantErr` + `errContain`.

For example:

```go
tests := []struct {
	name       string
	input      McpInput
	wantErr    string
	wantSecret []string
}{
	{
		name: "internal rejects empty secret name",
		input: McpInput{
			Port:       8080,
			ImageName:  "my-mcp",
			ImageTag:   "v1",
			SecretRefs: []string{"  "},
		},
		wantErr: `invalid --secret value "  "; name must not be empty`,
	},
}

for _, tt := range tests {
	t.Run(tt.name, func(t *testing.T) {
		body, err := BuildMcpRequestBody(tt.input)

		if tt.wantErr != "" {
			if err == nil || err.Error() != tt.wantErr {
				t.Fatalf("err = %v, want %q", err, tt.wantErr)
			}
			return
		}

		if err != nil {
			t.Fatalf("unexpected error: %v", err)
		}

		// success assertions...
		_ = body
	})
}
```

internal/inputs/mcp_create.go
```diff
@@ -0,0 +1,149 @@
+package inputs
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/Interactive-AI-Labs/interactive-cli/internal/clients"
+)
+
+// McpInput holds the flags for `iai mcps create`/`update`; the server does the real validation.
```
ajaen-ai: Nit: do we need a separate `mcp_create.go` here? `internal/inputs/mcp.go` is still small, and this builder is used by both create and update.

I’d probably keep `McpInput` / `BuildMcpRequestBody` in `mcp.go` and avoid the extra file unless this grows more.

ajaen-ai:
Review finished!

dgallego-iai:
@claude review again

cmd/agents.go
dgallego-iai: Kept singular to match --env/--secret/--header, which are also repeatable yet singular-named.

cmd/agents.go
dgallego-iai: Same reasoning as above — kept singular for consistency.

docs/iai_mcps_create.md
dgallego-iai: Removed the reference.

docs/iai_mcps_create.md
dgallego-iai: Reworded to drop the k8s-specific wording.

docs/iai_mcps_create.md
dgallego-iai: Condensed.

docs/iai_mcps_create.md
dgallego-iai: Reworded and dropped the pod reference.

docs/iai_mcps_get.md
dgallego-iai: Renamed to describe.

docs/iai_mcps_tools.md
dgallego-iai: Added — tools now show their argument names.

docs/iai_mcps_update.md
dgallego-iai: Added: the operator now has PATCH /mcps/{name} and iai mcps update uses it (see deployment-operator#230).

internal/output/mcps_test.go
dgallego-iai: Switched to exact expected-string comparison.

internal/inputs/mcp_create_test.go
dgallego-iai: Rewrote as a table-driven test with an exact want map, same as the rest of the repo.

internal/clients/deployment_client_mcps.go
dgallego-iai: Promoted to a generic sendJSONRequest on DeploymentClient; DescribeMcp now uses it too.

internal/inputs/mcp_create_test.go
dgallego-iai: Flattened to a single wantErr string as suggested.

internal/inputs/mcp_create.go
dgallego-iai: Merged mcp_create.go into mcp.go.

ajaen-ai (approved):

PR #163, opened 2026-07-09
feat(cli): DEV-799 — --watch/-w for list and describe

dgallego-iai:
## What

Adds `--watch`/`-w` to `iai services|agents|databases` **`list`** and **`describe`**. Polls every 2s and redraws in place so you can watch a rollout progress (healthy → progressing → healthy/rollback) without re-running the command.

Closes DEV-799.

## How

- New `cmd/watch.go`: `runWatch(cmd, render)` — `time.Ticker` loop that exits cleanly on Ctrl-C via `signal.NotifyContext` (same pattern as the `--follow` log commands), including a `ctx.Err()` guard so a Ctrl-C mid-request exits cleanly instead of printing `context canceled`.
- **Flicker-free redraw (no blink).** Instead of wiping the screen (`ESC[2J`) and then writing — which shows a blank frame each refresh — each frame homes the cursor (`ESC[H`), overwrites every line clearing its tail (`ESC[K`), then erases any leftover lines below (`ESC[J`). No blank frame between refreshes. `render` writes to a passed `io.Writer` so `runWatch` can post-process the frame (`redraw()`).
- Each `list`/`describe` RunE wraps its fetch+print in a `render` closure; `resolveProject` stays outside the loop.
- `-w` is mutually exclusive with `--json`/`--yaml`, and on `describe` with `--revision` (a pinned revision never changes).
- Non-TTY output (pipes) skips all escapes and just prints successive frames.
- No new dependencies.

## Scope note

The ticket says "list commands", but its motivating example (watch one agent you just updated transition) maps to `describe`, so `-w` was added there too — same helper, near-zero extra cost.

## Skipped (YAGNI)

- `--interval` flag — 2s hardcoded; trivial to add later.
- Keep-polling-through-transient-errors — aborts on error for v1.
- HTTP event stream — out of scope per the ticket.
- Home-based redraw tears if a frame is taller than the terminal; truncate to term height if that ever matters (flagged with a `ponytail:` comment).

## Test

- `go build ./cmd/iai`, `go vet ./...`, `go test ./cmd/...` green; `cmd/watch_test.go` covers cancel-stops-loop, cancel-during-render, error-propagation, and the `redraw()` transform.
- Live-verified against dev (`platform-development`/`dgallego`): list + describe refresh in place with no blink, revision transitions visible, clean Ctrl-C, no escape codes when piped; all mutual-exclusion errors fire.

### DIFF
--- a/cmd/watch.go
+++ b/cmd/watch.go
@@ -1,0 +1,53 @@
+package cmd
+
+import (
+	"bytes"
+	"context"
+	"io"
+	"os"
+	"os/signal"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/Interactive-AI-Labs/interactive-cli/internal/output"
+	"github.com/spf13/cobra"
+)
+
+const watchInterval = 2 * time.Second
+
+// runWatch re-runs render every watchInterval until Ctrl-C, redrawing in place.
+func runWatch(cmd *cobra.Command, render func(context.Context, io.Writer) error) error {
+	ctx, stop := signal.NotifyContext(cmd.Context(), os.Interrupt, syscall.SIGTERM)
+	defer stop()
+
+	ticker := time.NewTicker(watchInterval)
+	defer ticker.Stop()
+
+	out := cmd.OutOrStdout()
+	tty := output.IsTerminal(out)
+	for {
+		var frame bytes.Buffer
+		if err := render(ctx, &frame); err != nil {
+			if ctx.Err() != nil {
+				return nil // Ctrl-C during an in-flight request
+			}
+			return err
+		}
+		if tty {
+			io.WriteString(out, "\033[H"+redraw(frame.String()))
+		} else {
+			out.Write(frame.Bytes())
+		}
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-ticker.C:
+		}
+	}
+}
+
+// redraw overwrites the previous frame in place so there is no blank flash between refreshes.
+func redraw(frame string) string {
+	return strings.ReplaceAll(frame, "\n", "\033[K\n") + "\033[J"
+}
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -1,5 +1,5 @@
 [tool.bumpversion]
-current_version = "0.37.1"
+current_version = "0.37.2"
 commit = true
 tag = false
 parse = "(?P<major>\\d+)\\.(?P<minor>\\d+)\\.(?P<patch>\\d+)(-beta\\.(?P<dev>\\d+))?"
--- a/cmd/agents.go
+++ b/cmd/agents.go
@@ -1,7 +1,9 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
+	"io"
 	"os"
 	"os/signal"
 	"strings"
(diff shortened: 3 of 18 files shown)


dgallego-iai:
Thanks for the review. Addressed the blocking bug in 2ce21cc… (latest push):

- **Ctrl-C mid-request** — added the `ctx.Err() != nil → return nil` guard in `runWatch`, matching the log-follow pattern you pointed to. Now exits cleanly instead of printing `context canceled`.
- **Missing test** — added `TestRunWatchCleanExitOnCancelDuringRender` covering a render that returns `context.Canceled` because the context was cancelled.

Non-blocking notes left as-is on purpose:
- `isTerminal` only handling `*os.File` — fine for the default Cobra stdout path; not worth generalizing (YAGNI).
- `ponytail:` tags — deliberate convention marking intentional simplifications (hardcoded interval, abort-on-error) with their upgrade path.

cmd/watch.go
```diff
@@ -0,0 +1,61 @@
...
+	"os/signal"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/spf13/cobra"
+	"golang.org/x/term"
+)
+
+const watchInterval = 2 * time.Second // ponytail: hardcoded; add --interval when asked
```
ajaen-ai: Lets not add "ponytail" comments.

Apart from that, I think the comment is not necessary.

cmd/watch.go
```diff
@@ -0,0 +1,61 @@
...
+
+	out := cmd.OutOrStdout()
+	tty := isTerminal(out)
+	for {
+		var frame bytes.Buffer
+		if err := render(ctx, &frame); err != nil {
+			if ctx.Err() != nil {
+				return nil // Ctrl-C during an in-flight request
+			}
+			return err // ponytail: abort on error; retry through transient errors when asked
```
ajaen-ai: Same here

cmd/watch.go
```diff
@@ -0,0 +1,61 @@
...
+		case <-ctx.Done():
+			return nil
+		case <-ticker.C:
+		}
+	}
+}
+
+// redraw overwrites the previous frame in place instead of wiping first: each
+// line clears its own tail and any leftover lines below are erased, so there is
+// no blank flash between frames. ponytail: home-based, tears if the frame is
```
ajaen-ai: Ponytail again.

Apart from that, it looks like there is too much detail in the complete comment.

cmd/watch.go
```diff
@@ -0,0 +1,61 @@
...
+// taller than the terminal; truncate to term height if that ever matters.
+func redraw(frame string) string {
+	return strings.ReplaceAll(frame, "\n", "\033[K\n") + "\033[J"
+}
+
+func isTerminal(w io.Writer) bool {
+	f, ok := w.(*os.File)
+	return ok && term.IsTerminal(int(f.Fd()))
+}
```
ajaen-ai: I think we have this func already elsewhere in the code

cmd/watch.go
dgallego-iai: Done — removed the comment.

cmd/watch.go
dgallego-iai: Done — removed.

cmd/watch.go
dgallego-iai: Done — dropped the extra detail, trimmed to a single line.

cmd/watch.go
dgallego-iai: Good catch — reused it. Exported the existing `isTerminal` in `internal/output/logs.go` as `IsTerminal` and dropped the duplicate here.

ajaen-ai:
Approved, remember to squash the commits into the shape that makes sense to have a clean git history!

ajaen-ai (approved):

PR #164, opened 2026-07-10
Fix followed logs max connection handling

dgallego-iai:
## Summary
- add a 10-minute deadline for followed log streams
- print a clean max-connection-time message instead of surfacing transport errors
- share the behavior across service, agent, database, and replica logs

## Validation
- `go test ./...`
- `go vet ./...`
- local dev test: `agents logs -f` on installed CLI reproduced `stream error: stream ID 1; INTERNAL_ERROR; received from peer`
- local fixed binary closed after 600s with `Maximum connection time is 10 minutes, closing connection.`

### DIFF
--- a/cmd/agents.go
+++ b/cmd/agents.go
@@ -2,10 +2,7 @@ package cmd
 
 import (
 	"fmt"
-	"os"
-	"os/signal"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/Interactive-AI-Labs/interactive-cli/internal/clients"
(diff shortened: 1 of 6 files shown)


dgallego-iai:
Addressed Claude's minor note in 434686f: the max connection message now derives from `logFollowMaxConnectionTime.Minutes()` instead of hard-coding "10 minutes".\n\nValidation: `go test ./...` passes.

dgallego-iai:
Closing this CLI-side workaround after tracing the root cause. Deployment operator v0.45.7 already emits a clean message at its 10-minute stream deadline. Agent and database streams were instead falling through to the generic gateway route and being reset after ~5 minutes; existing service/replica routes also raced the operator at exactly 10 minutes. The infrastructure fix is Interactive-AI-Labs/infra-k8s#1724.

PR #123, opened 2026-05-14
Add port-forwarding for services agents and dbs

ajaen-ai:

PR #166, opened 2026-07-15
fix(DEV-857): use the active label in skills help and docs

gabrielgustavoandrade:
## What

Update the `iai skills` command help text, examples, and generated docs to reference the `active` label instead of `production`.

- `skills create` / `skills update`: examples now use `--labels active`.
- `skills create` long help: the `active` label marks the version Copilot loads; drop the guidance to label skills `production`.
- `skills get` long help: `--label active` fetches the version Copilot uses. The bare `get` default description stays accurate — the server resolves the `production` label if present, otherwise the highest version.
- `skills get` example: `--label staging` → `--label active`.
- Regenerated `docs/iai_skills_*.md` via `make docs`.

## Why

Platform (DEV-857) retired the built-in Langfuse `production` prompt label for Copilot skills. Copilot now resolves skills via `label=active`, loading the `active` version of each skill (platform PR #1245, copilot PR #123). The `production` label still exists as an ordinary user label for other prompt types, and the server's default label-less resolution is unchanged — so only the skills help text and examples needed updating. No hardcoded label logic exists in the CLI.

Skills-only; other command help (prompts, routines, policies, macros, variables) is untouched.

### DIFF
--- a/cmd/skills.go
+++ b/cmd/skills.go
@@ -36,13 +36,13 @@ Example (skill.md):
   Given a Langfuse trace ID, fetch the trace and summarize key steps,
   latencies, and any errors.
 
-The server automatically assigns the "latest" label to new versions. To make
-a version retrievable via the default 'get' (which resolves "production"),
-assign the "production" label with --labels production.`,
+The server automatically assigns the "latest" label to new versions. Copilot
+loads the version labeled "active", so assign it with --labels active to make
+a skill the one Copilot uses.`,
 		CreateExample: `  iai skills create summarize-trace --file ./skill.md \
     --description "Summarize a Langfuse trace" \
     --intents "summarize trace" --intents "explain trace"
-  iai skills create summarize-trace --file ./skill.md --labels production`,
+  iai skills create summarize-trace --file ./skill.md --labels active`,
 		ListLong: `List Copilot skills in a project.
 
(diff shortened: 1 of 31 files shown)


dgallego-iai (approved):

cmd/skills.go
```diff
@@ -53,11 +53,14 @@ can be browsed into with --folder.`,
   iai skills list --page 2 --limit 10`,
 		GetLong: `Show a Copilot skill in detail, including its config and full content.
 
-By default returns the version labeled "production". Use --version to retrieve a
-specific version number, or --label to resolve a different label.`,
+A label-less get returns whatever the server resolves by default: the version
+labeled "production" if one exists, otherwise the highest version number.
```
ajaen-ai: Haven't we completely deleted the special treatment of the "production" label?

gabrielgustavoandrade: Good catch — the CLI passes labels through verbatim and shouldn't editorialize about server resolution at all. Dropped the production-default explanation here and removed the remaining "resolves 'production'" claims from the shared --label flag help and the other prompt-type long-help texts in 561d532. Examples now use generic labels (staging, staging,qa).

ajaen-ai (approved):

ppalacio-iai (approved):

PR #169, opened 2026-07-17
feat: add router command and nest model commands

dgallego-iai:
## Summary

- add `iai router` as the parent for router discovery and model commands
- add `iai router info` with human, `--json`, and `--yaml` endpoint output derived from `--hostname`
- move `iai models list/get` to `iai router models list/get`
- regenerate CLI documentation for the new command paths

## Commands

```bash
iai router                          # parent help
iai router info                     # human-readable endpoint details
iai router info --json              # machine-readable endpoint details
iai router info --yaml
iai router models list
iai router models get <id>
```

## UX decision

`iai router` keeps Cobra's default parent behavior and prints help. `iai router info` provides endpoint URLs in human- and machine-readable formats.

## Breaking change

The top-level `iai models` path is removed. Its replacement is `iai router models`.

## Version

Rebased onto `main` at `v0.38.1` and bumped to `v0.38.2` after approval. The tag will be created automatically after merge.

## Verification

- `pre-commit run --from-ref origin/main --to-ref HEAD`
- `go test ./...`
- `make docs`
- `go run . router`
- `go run . router info --hostname https://dev.interactive.ai --json`
- `go run . router models list --hostname https://dev.interactive.ai --deployment-hostname https://deployment.dev.interactive.ai -o platform-development -p dgallego --limit 1 --json`

### DIFF
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -1,5 +1,5 @@
 [tool.bumpversion]
-current_version = "0.38.1"
+current_version = "0.38.2"
 commit = true
 tag = false
 parse = "(?P<major>\\d+)\\.(?P<minor>\\d+)\\.(?P<patch>\\d+)(-beta\\.(?P<dev>\\d+))?"
--- a/cmd/models.go
+++ b/cmd/models.go
@@ -30,7 +30,6 @@ var modelsCmd = &cobra.Command{
 	Use:              "models",
 	Aliases:          []string{"model"},
 	Short:            "List and inspect models",
-	GroupID:          groupInfra,
 	Long:             `List and inspect router models available to a project.`,
 	PersistentPreRun: chainRootPersistentPreRun,
 }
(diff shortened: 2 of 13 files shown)


docs/iai_router_get.md
```diff
@@ -0,0 +1,42 @@
+## iai router get
```
ajaen-ai: Didn't we agree on info? I think it is more understandable than get

internal/clients/router.go
ajaen-ai, line 1: This doesn't connect to anything right? Clients are our representation of remote servers that we connect to, I would add this code elsewhere (outputs?)

internal/clients/router_test.go
ajaen-ai, line 1: This is pure hardcoded values right? I wouldn't add tests for this

docs/iai_router_get.md
dgallego-iai: Was thinking about that, `info` doesn't follow either our existing convention or the major CLIs' conventions.

I was between `get` or `describe` but `info` could fit as well. I don't have an strong opinion on this one

internal/clients/router.go
dgallego-iai: Moved the router metadata and URL construction to `internal/output/router.go`; the command no longer adds a non-remote client.

internal/clients/router_test.go
dgallego-iai: Removed the duplicated client test with the hardcoded endpoint values. The remaining output tests cover the user-facing human/JSON contract, and a small table covers hostname validation.

docs/iai_router_get.md
dgallego-iai: Updated the command to `iai router info` and regenerated its docs.

ajaen-ai (approved):

PR #172, opened 2026-07-21
feat(mcps): surface a tool's JSON-RPC error from run-tool and exit non-zero (DEV-840)

dgallego-iai:
## Context

Pairs with deployment-operator PR #232. The operator now treats a JSON-RPC error from a **reached** MCP as a completed call whose result is an error: it returns **HTTP 200** with an `error` {code, message} object (the transport succeeded — the tool reported an error), rather than inventing an HTTP status. Genuine transport failures (MCP unreachable, timeout, non-2xx) return 424.

## Change

- `RunMcpToolResult` gains an `Error *McpToolError` field to decode the 200-with-error body.
- `iai mcps run-tool`: when the body carries an `error`, print a clear message naming the mcp, tool, and JSON-RPC code, and **return non-zero** so scripts using `&&` don't treat a tool failure as success. Success still prints the raw result as before.

Example:
```
$ iai mcps run-tool socket depscore --args '{"packages":[{"name":"lodash"}]}'
Error: mcp "socket" tool "depscore" returned an error (JSON-RPC -32603): Cannot read properties of undefined (reading 'startsWith')
$ echo $?
1
```

## Testing

- Pure decode test (`TestDecodeRunMcpToolResult`) covering both the error body and the success body — confirms the struct tags route correctly. No mocks.
- Verified against the operator change live: Socket `depscore` bad args → `-32603` in a 200 → this path; Socket `organizations` → real 401 → the existing 424 error path.

Requires operator PR #232 (returns the 200-with-error shape) to be deployed for the new path to trigger; until then run-tool behaves as today.

### DIFF
--- a/cmd/mcps.go
+++ b/cmd/mcps.go
@@ -481,8 +481,8 @@ var mcpRunToolCmd = &cobra.Command{
 	Long: `Call one of an mcp's tools and print the result.
 
 Pass arguments as a JSON object with --args or --args-file (mutually exclusive);
-omit both to send an empty object. Works for external mcps from anywhere;
-internal mcps need the in-cluster operator.`,
+omit both to send an empty object. If the tool itself returns an error, it is
+reported and the command exits non-zero.`,
 	Example: `  iai mcps run-tool github search_repositories --args '{"query":"interactiveai"}'
   iai mcps run-tool github search_repositories --args-file ./args.json`,
 	Args: cobra.ExactArgs(2),
@@ -506,6 +506,12 @@ internal mcps need the in-cluster operator.`,
 		)
 		if err != nil {
 			return err
+		}
+		if res.Error != nil {
+			return fmt.Errorf(
+				"mcp %q tool %q returned an error (JSON-RPC %d): %s",
+				mcpName, tool, res.Error.Code, res.Error.Message,
+			)
 		}
 		return output.PrintRawJSON(out, res.Result)
 	},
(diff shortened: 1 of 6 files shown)


cmd/mcps.go
```diff
@@ -482,7 +482,8 @@ var mcpRunToolCmd = &cobra.Command{
 
 Pass arguments as a JSON object with --args or --args-file (mutually exclusive);
 omit both to send an empty object. Works for external mcps from anywhere;
-internal mcps need the in-cluster operator.`,
+internal mcps need the in-cluster operator. If the tool itself returns an error,
```
ajaen-ai: I would delete "internal mcps need the in-cluster operator"

internal/clients/api_client_mcp_test.go
```diff
@@ -1,9 +1,24 @@
...
+	if err := json.Unmarshal(body, &res); err != nil {
+		t.Fatalf("decode error: %v", err)
+	}
+	if res.Error == nil || res.Error.Code != -32603 || res.Error.Message != "boom" {
+		t.Fatalf("unexpected Error: %#v", res.Error)
+	}
+	if len(res.Result) != 0 {
+		t.Fatalf("expected empty Result, got %s", res.Result)
+	}
+}
```
ajaen-ai, line 44: Tabled based

cmd/mcps.go
dgallego-iai: Done — removed that clause.

internal/clients/api_client_mcp_test.go
dgallego-iai: Done — converted it to a table-driven test.

ajaen-ai (approved):

PR #175, opened 2026-07-27
feat(DEV-877): surface recommendation ranks in router models output

gabrielgustavoandrade:
## What

`iai router models list` / `get` never showed the per-category recommendation rank, even though the API has returned it since DEV-877.

`GET /api/v1/models/router-models` returns `recommendations: {"chat": 3, "vision": 3}` (effective rank, `COALESCE(override_rank, seeded_rank)`), but `clients.RouterModel` had no matching field — `json.Unmarshal` silently dropped it. `--json`/`--yaml` were affected too, since both re-marshal the struct rather than passing the raw body through.

Verified against dev before writing the fix: 19 of 32 models on `eu.dev.interactive.ai/api/v1/models` carry a non-empty `recommendations` map, so the data was there the whole time.

## Changes

- `internal/clients/router_models.go` — add `Recommendations map[string]int` to `RouterModel`. This alone fixes `--json` and `--yaml`.
- `internal/output/models.go` — `RECOMMENDED` column in the list table, `Recommended:` line in the detail view, both via `formatRecommendations` (renders `chat:1 vision:1`, category-sorted so output is stable; `-` when empty).
- `internal/output/models_test.go` — updated the golden list output for the new column, plus a case covering the sorted rendering and a detail-view assertion.

## Testing

`go build ./...`, `go vet ./...`, `gofmt -l` clean, `go test ./internal/...` passing.

Not smoke-tested against a live project — I'm not authenticated locally. The decode path is a plain tag match on a field confirmed present in the live dev response.

No version bump here; recent feature PRs land without one and `.bumpversion.toml` auto-tags on its own release commit.

### DIFF
--- a/internal/output/models_test.go
+++ b/internal/output/models_test.go
@@ -97,3 +97,16 @@ func TestPrintRouterModelDetail(t *testing.T) {
 		}
 	}
 }
+
+// The detail view omits the line entirely rather than printing the list
+// view's "-" placeholder.
+func TestPrintRouterModelDetailOmitsEmptyRecommendations(t *testing.T) {
+	var buf bytes.Buffer
+	model := &clients.RouterModel{ID: "m-1", ModelName: "gpt-4o", Region: "us"}
+	if err := PrintRouterModelDetail(&buf, model); err != nil {
+		t.Fatalf("PrintRouterModelDetail() error = %v", err)
+	}
+	if strings.Contains(buf.String(), "Recommended:") {
+		t.Errorf("expected no Recommended line\ngot:\n%s", buf.String())
+	}
+}
--- a/internal/output/models_test.go
+++ b/internal/output/models_test.go
@@ -69,42 +69,57 @@ func TestPrintRouterModelList(t *testing.T) {
 }
 
 func TestPrintRouterModelDetail(t *testing.T) {
-	var buf bytes.Buffer
-	model := &clients.RouterModel{
-		ID:              "m-1",
-		ModelName:       "gpt-4o",
-		MatchPattern:    "gpt-4o*",
-		Region:          "us",
-		Capabilities:    []string{"text", "vision"},
-		Prices:          map[string]float64{"input": 0.01, "output": 0.03},
-		Recommendations: map[string]int{"chat": 3, "vision": 3},
+	tests := []struct {
+		name        string
+		model       clients.RouterModel
+		contains    []string
+		notContains []string
+	}{
+		{
+			name: "recommended model names each category",
+			model: clients.RouterModel{
+				ID:              "m-1",
+				ModelName:       "gpt-4o",
+				MatchPattern:    "gpt-4o*",
+				Region:          "us",
+				Capabilities:    []string{"text", "vision"},
+				Prices:          map[string]float64{"input": 0.01, "output": 0.03},
+				Recommendations: map[string]int{"chat": 3, "vision": 3},
+			},
+			contains: []string{
+				"ID:             m-1",
+				"Model Name:     gpt-4o",
+				"Capabilities:   text, vision",
+				"Recommended:    chat #3, vision #3",
+				"Prices:",
+				"  input:    0.01",
+				"  output:   0.03",
+			},
+		},
+		{
+			name:        "unrecommended model drops the line rather than printing the list view's placeholder",
+			model:       clients.RouterModel{ID: "m-1", ModelName: "gpt-4o", Region: "us"},
+			notContains: []string{"Recommended:"},
+		},
 	}
-	if err := PrintRouterModelDetail(&buf, model); err != nil {
-		t.Fatalf("PrintRouterModelDetail() error = %v", err)
-	}
-	got := buf.String()
-	for _, want := range []string{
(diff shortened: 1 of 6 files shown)


gabrielgustavoandrade:
Both nits applied in 47a4516.

**Field ordering** — `Recommended:` now sits right after `Other Region:`, before `Description`. Agreed it reads better grouped with the other high-level attributes than wedged between `Last Used` and `Prices`.

**Empty-case coverage** — added `TestPrintRouterModelDetailOmitsEmptyRecommendations`. The asymmetry was deliberate (list view prints `-` to keep the column aligned, detail view drops the line), so it's worth locking in. Comment on the test says as much.

`gofmt`, `go vet`, and `go test ./internal/...` all clean.

@claude please re-review.

gabrielgustavoandrade:
Ran the built binary end to end, not just the unit tests. I'm not authenticated against dev locally, so I pointed it at a local stub serving the live dev response shape (`--hostname`/`--api-key`, which resolves org/project off the `validate-api-key` headers). Fixture covers all four categories plus one unrecommended model.

`iai router models list`:

```
NAME                                        CONTEXT   REGION   RECOMMENDED          ID
interactive/anthropic/claude-opus-5         200000    eu       chat #1, vision #1   m-1
interactive/anthropic/claude-sonnet-5       200000    eu       chat #5, vision #5   m-2
interactive/openai/text-embedding-3-large   8191      us       embedding #1         m-3
interactive/cohere/rerank-v3.5              -         us       reranking #2         m-4
interactive/mistral/mistral-small           32768     eu       -                    m-5
```

`iai router models get m-1`:

```
ID:             m-1
Model Name:     interactive/anthropic/claude-opus-5
Match:          claude-opus-5*
Region:         eu
Other Region:   Yes
Recommended:    chat #1, vision #1
Context:        200000
Capabilities:   Text, Vision, Tools
```

`--json` carries the map through, and `omitempty` drops it on the unrecommended row: `[('m-1', {'chat': 1, 'vision': 1}), ..., ('m-5', '<absent>')]`.

Same stub against a binary built from `main` prints the old four-column table with no rank anywhere, which is the bug this fixes.

Also added the decode assertions that were actually missing — `TestAPIClientListRouterModels` and `TestAPIClientGetRouterModelByID` now put `recommendations` in the stub payload and assert it survives unmarshal (`ef02e07`). That's the test that would have caught the original drop; the output tests only covered formatting.

@claude please re-review.

gabrielgustavoandrade:
@ajaen-ai adding you as a reviewer because Daniel is out

internal/output/models_test.go
```diff
@@ -79,3 +97,14 @@ func TestPrintRouterModelDetail(t *testing.T) {
...
+func TestPrintRouterModelDetailOmitsEmptyRecommendations(t *testing.T) {
+	var buf bytes.Buffer
+	model := &clients.RouterModel{ID: "m-1", ModelName: "gpt-4o", Region: "us"}
+	if err := PrintRouterModelDetail(&buf, model); err != nil {
+		t.Fatalf("PrintRouterModelDetail() error = %v", err)
+	}
+	if strings.Contains(buf.String(), "Recommended:") {
+		t.Errorf("expected no Recommended line\ngot:\n%s", buf.String())
+	}
+}
```
ajaen-ai: Can we make this table based like we do in the rest of the repo?

ajaen-ai:
Could we cover this with a table-driven JSON input/output test instead? `httptest.NewServer` is a fake HTTP server, which our unit-testing guidance avoids. Since both endpoints decode into `RouterModel`, directly unmarshalling cases with present, missing, and invalid `recommendations` verifies the contract once without mocking transport. We’ll add integration tests against the real service stack separately to cover HTTP request/response behavior.

ajaen-ai:
Review finished!

internal/output/models_test.go
gabrielgustavoandrade: Done in af929cf — folded it into `TestPrintRouterModelDetail` as a two-case table (`contains`/`notContains`), matching `TestPrintRouterModelList` right above it. The standalone test is gone.

gabrielgustavoandrade:
Both addressed in af929cf.

**Table-driven decode instead of `httptest`** — agreed, and the direct unmarshal is the better test anyway: the bug was a missing struct tag, which has nothing to do with transport. Reverted my additions to `TestAPIClientListRouterModels` / `TestAPIClientGetRouterModelByID` (they're back to exactly what they were before this PR — I'd only bolted assertions onto the existing fake servers, which was the wrong place for them) and added `TestRouterModelUnmarshalRecommendations` in `internal/clients/router_models_test.go`, covering present / absent / null / empty-object / non-numeric-rank. The last case pins that a bad rank type is a decode error rather than a silent zero.

Makes sense to leave real HTTP behaviour to the integration tests against the service stack.

**Detail-view test** — folded into `TestPrintRouterModelDetail` as a two-case table, same shape as `TestPrintRouterModelList` above it. Replied inline too.

`gofmt`, `go vet`, and `go test ./...` clean.

@claude please re-review.

ajaen-ai (approved):

PR #176, opened 2026-07-27
feat(deploy): add deploy awareness warnings

oryall:
## Summary

Deploy-mutating commands now surface the live state they are about to replace, and **refuse** destruction that is only implied by omission until an explicit flag names the intent.

| Operation | What gets guarded | Default | Override |
|---|---|---|---|
| `agents/services update` | pin downgrade / removal / unpin; `--env`/`--secret` drops of live names | refuse | `--force` |
| `images push` | overwrite of an existing upstream tag | refuse | `--force` |
| `stacks sync` / `services sync` | delete of services / agents / databases omitted from the config | refuse (creates/updates still apply) | `--allow-delete=<types\|all>` |
| `agents/services update` | live revision differs from expected | warn/proceed (banner only) | `--expect-revision N` to fail closed |
| any of the above | live state cannot be fetched | **fail open** (note on stderr) | — |

Also:
- stderr banner with live revision + last-updated before agent/service updates and sync replacements
- content-pin delta summary; opt-in `--show-diff` for a full live-vs-incoming config diff
- `stacks sync --dry-run` prints the full plan (creates, updates, deletes, refused deletions with rationale) without writing

## Why / decisions

Agent configs and image tags are mutable deployment slots. A stale local manifest can silently revert another actor's pins, wipe env/secrets via replace-not-merge flags, delete resources simply by omitting them from a sync file, or irrecoverably replace image bytes under a reused tag.

**Decision: fail closed on omission-implied destruction, never prompt.** Warnings alone are invisible to LLM copilots until after the mutation. A single explicit flag (`--force` or `--allow-delete`) is the unblock path so the first attempt fails with details on stderr and a deliberate second command applies.

**Decision: fail open when live state cannot be fetched.** Gates must not invent a new way for a legitimate deploy to break. `--expect-revision` is the opt-in exception (it cannot proceed without the live revision).

**Decision: warn vs gate by pin section.** Known sections (`system_prompt`, `policies`, `routines`, `glossaries`, `macros`) gate on downgrade/removal/unpin. Unknown `{id, version}`-shaped context sections warn only — never block on guessed semantics.

**Decision: `--clear-env` / `--clear-secret` stay silent.** Clearing is explicit intent; only full-list replacements that quietly drop live names gate.

## Behavior notes

- All deploy-awareness output is deterministic plain text on **stderr** (`⚠` prefix on warnings). Stdout payloads and non-gated exit codes stay unchanged.
- Sync deletions, when allowed, still run **after** that resource type's creates/updates; the refusal/allow announcement prints first so there is an abort window.
- Dry-run "Would refuse to delete" lines carry the same stale-config rationale and `--allow-delete` hint as a live run's stderr warning.
- `--expect-revision 0` is valid (matches a never-updated resource); the gate keys off the flag being set, not a non-zero value.
- `--expect-revision` is a client-side check, not atomic compare-and-swap.
- One agent `Describe` feeds the banner, pin summary, `--show-diff`, `--expect-revision`, and MCP overlay.

## Implementation

- `internal/preflight` — banners, pin comparison, tag warnings, sync deletion announcements, dropped list warnings, fail-open notes, `--expect-revision`
- `cmd/preflight.go` — shared update wiring (`runUpdatePreflight`, gates, env/secret drop checks, config preflight)
- `internal/sync` — generic `syncResources` + `resourceOps` adapters for services/agents/databases; `Options{AllowDelete, DryRun}`; `PrintPlan` for dry-run
- `cmd/stack.go` — single `runPhase` helper for progress + apply/plan printing
- `cmd/images.go` — tag-overwrite gate after cheap local checks; reuses the push command's `DeploymentClient`
- docs updated for `agents update`, `services update`, `images push`, `stacks sync`, `services sync`

## Compatibility

**Breaking for previously silent destructive paths:**
- sync no longer deletes omitted services/agents without `--allow-delete` (databases already behaved this way)
- updates that downgrade/remove pins or drop live env/secret names now error without `--force`
- pushing to an existing image tag now errors without `--force`

Non-breaking:
- revision banners, pin summaries, and fail-open notes are stderr-only
- default update/sync success paths that do not destroy by omission are unchanged

## Testing

- [x] `uvx pre-commit run --all-files`
- [x] `go vet ./...`
- [x] `TZ=Europe/Berlin go test -count=1 ./...`
- [x] command-level tests: MCP overlay `--show-diff` regression, unpinned routine refused, dropped `--env` refused
- [x] stub-API sync tests: deletion announced before writes; refuse-by-default and `--allow-delete` paths for services/agents/databases
- [x] stub-API scenario replay for banners, pin downgrades, tag reuse, `--expect-revision`, fail-open
- [x] dry-run plan includes refused-deletion rationale

### DIFF
--- a/cmd/images_test.go
+++ b/cmd/images_test.go
@@ -85,7 +85,7 @@ func TestWarnExistingImageTag(t *testing.T) {
 			status:     http.StatusInternalServerError,
 			imagesJSON: `{}`,
 			tag:        "0.2.36",
-			wantPrefix: "could not list existing image tags (",
+			wantPrefix: "⚠ could not list existing image tags (",
 		},
 	}
 
--- a/cmd/images_test.go
+++ b/cmd/images_test.go
@@ -50,41 +50,42 @@ func TestImageDeleteValidation(t *testing.T) {
 	}
 }
 
-func TestWarnExistingImageTag(t *testing.T) {
+func TestExistingImageTag(t *testing.T) {
 	tests := []struct {
 		name       string
 		status     int
 		imagesJSON string
 		tag        string
-		want       string
+		wantExists bool
 		wantPrefix string
 	}{
 		{
-			name:       "existing tag warns",
+			name:       "existing tag detected",
 			status:     http.StatusOK,
 			imagesJSON: `{"images":[{"name":"app","tags":["0.2.35","0.2.36"]}]}`,
 			tag:        "0.2.36",
-			want:       "⚠ tag 0.2.36 already exists upstream — pushing replaces it; the previous image is unrecoverable.\n",
+			wantExists: true,
 		},
 		{
-			name:       "new tag stays silent",
+			name:       "new tag reports absent",
 			status:     http.StatusOK,
 			imagesJSON: `{"images":[{"name":"app","tags":["0.2.35"]}]}`,
 			tag:        "0.2.36",
-			want:       "",
+			wantExists: false,
 		},
 		{
-			name:       "same tag on another image stays silent",
+			name:       "same tag on another image reports absent",
 			status:     http.StatusOK,
 			imagesJSON: `{"images":[{"name":"other","tags":["0.2.36"]}]}`,
 			tag:        "0.2.36",
-			want:       "",
+			wantExists: false,
 		},
 		{
 			name:       "list failure fails open with a note",
 			status:     http.StatusInternalServerError,
 			imagesJSON: `{}`,
 			tag:        "0.2.36",
+			wantExists: false,
 			wantPrefix: "⚠ could not list existing image tags (",
 		},
 	}
@@ -104,8 +105,11 @@ func TestWarnExistingImageTag(t *testing.T) {
 			deploymentHostname, token = server.URL, "test-token"
 
 			var buf bytes.Buffer
-			warnExistingImageTag(context.Background(), &buf, "o1", "p1", "app", tt.tag, nil)
+			exists := existingImageTag(context.Background(), &buf, "o1", "p1", "app", tt.tag, nil)
 
+			if exists != tt.wantExists {
+				t.Errorf("exists = %v, want %v", exists, tt.wantExists)
+			}
 			got := buf.String()
 			if tt.wantPrefix != "" {
 				if !strings.HasPrefix(got, tt.wantPrefix) ||
@@ -114,8 +118,11 @@ func TestWarnExistingImageTag(t *testing.T) {
 				}
 				return
 			}
-			if got != tt.want {
-				t.Errorf("output = %q, want %q", got, tt.want)
+			if got != "" {
+				t.Errorf(
+					"output = %q, want none (the caller decides whether to warn or refuse)",
+					got,
+				)
 			}
 		})
 	}
--- a/cmd/images_test.go
+++ b/cmd/images_test.go
@@ -1,6 +1,12 @@
 package cmd
 
 import (
+	"bytes"
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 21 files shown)


oryall:
@claude I have made some structural changes to the deploy-guard code and refactored, Give me another review

dgallego-iai:
@oryall Could you please trim the comments across this PR. We try to keep stick to the repo rule of keeping comments short (ideally one line) and use them only when really needed

internal/preflight/preflight.go
```diff
@@ -0,0 +1,345 @@
...
+					addPin(pins, label, known, m)
+				}
+			}
+		}
+	}
+	return pins
+}
+
+func addPin(pins map[pinKey]pinVersion, label string, known bool, entry map[string]any) {
+	id, _ := entry["id"].(string)
```
dgallego-iai, line 179: Older supported agent configs use `context.description.prompt_id`, but this only reads `id`. A stale file can therefore roll back that prompt without warning. Please support this format and add a table test.

dgallego-iai:
Could you please update the PR description with the latest changes? It will make the final behavior and decisions easier to track later.

internal/sync/sync.go
```diff
@@ -285,8 +290,68 @@ func Databases(
...
+	banner    func(w io.Writer, existing E)
+}
+
+// syncResources applies the shared sync contract for one resource type:
+// announce the deletion decision on warnW before any write, refuse deletions
+// without opts.AllowDelete (reporting them as Protected), create/update the
+// desired resources in name order, delete last, and — with opts.DryRun —
+// accumulate the same plan into the Result without calling the API at all.
+func syncResources[E, B any](
+	ctx context.Context,
```
dgallego-iai: `ctx` is not used inside `syncResources`; the resource functions already capture it. Please remove this parameter.

internal/sync/sync_test.go
```diff
@@ -163,3 +169,453 @@ func TestPrintResult(t *testing.T) {
...
+	}
+	if len(result.Updated) != 1 || result.Updated[0] != "svc-a" {
+		t.Errorf("Updated = %v, want [svc-a]", result.Updated)
+	}
+	if len(result.Created) != 1 || result.Created[0] != "svc-new" {
+		t.Errorf("Created = %v, want [svc-new]", result.Created)
+	}
+}
+
+func TestServicesRefusesDeletionsByDefault(t *testing.T) {
```
dgallego-iai: These deletion cases repeat the same setup and checks for each resource. Could we make them table-driven with the resource name, options, and expected result? That would keep the shared sync behavior covered with less repeated test code.

cmd/images_test.go
```diff
@@ -43,3 +50,86 @@ func TestImageDeleteValidation(t *testing.T) {
 		})
 	}
 }
+
+func TestExistingImageTag(t *testing.T) {
```
dgallego-iai, line 54: This tests the tag lookup, but not the push behavior that depends on it. Could we add table cases showing that an existing tag stops the command without `--force`, and that `--force` allows it to continue?

internal/preflight/preflight.go
```diff
@@ -0,0 +1,345 @@
...
+
+// PrintFailOpenNote reports that a pre-flight lookup failed and the command
+// is proceeding without it. what describes the lookup, e.g. "fetch live state".
+func PrintFailOpenNote(w io.Writer, what string, err error) {
+	fmt.Fprintf(w, "⚠ could not %s (%v) — proceeding without pre-flight check\n", what, err)
+}
+
+// PrintTagOverwriteWarning warns that pushing to an already-existing tag
+// replaces the previous image with no way to recover it.
+func PrintTagOverwriteWarning(w io.Writer, tag string) {
```
dgallego-iai: This is a one-line function with one caller. Could we print the message at the call site instead? The extra function and isolated test add another place to follow without providing reuse.

internal/preflight/preflight.go
```diff
@@ -0,0 +1,345 @@
...
+		kind,
+		strings.Join(dropped, ", "),
+		flag,
+	)
+	return true
+}
+
+// CheckExpectedRevision enforces --expect-revision: the update must not be
+// applied when the live revision differs from what the caller expects.
+func CheckExpectedRevision(expected, live int) error {
```
dgallego-iai: This helper has one caller and only contains a single comparison. Could we move the check into `runUpdatePreflight` so the revision decision is visible in one place?

internal/sync/sync.go
```diff
@@ -295,52 +360,53 @@ func Databases(
...
 		}
+		result.Deleted = append(result.Deleted, name)
 	}
 
 	return result, nil
 }
+
+// deletionList returns, sorted, the existing resource names a sync will
+// delete because the desired config no longer mentions them.
+func deletionList[E, D any](existing map[string]E, desired map[string]D) []string {
```
dgallego-iai: Same here, this helper has one caller and contains one short loop. Could we move it into `syncResources` to keep the deletion logic together?

dgallego-iai (approved):
Tested locally against DEV. The requested review comments were addressed, and the safety checks behave as expected. Approved.

dgallego-iai:
@oryall One final repo note about version bumps: please bring the branch up to date with `main` first (`main` is currently 0.38.7 while this branch is 0.38.6), then run `bump-my-version bump patch`. It updates `.bumpversion.toml` and `cmd/root.go` and creates the bump commit. Do not create the tag manually; after merge, the auto-tag workflow will create `v0.38.8`.

PR #177, opened 2026-08-03
fix(traces): repair turn summaries and add batch summaries to traces list

oryall:
## Summary

Trace investigation through the CLI had quietly degraded: `iai traces get --summary` — the compact view of what an agent did in a turn — rendered **"0 iterations"** with no tools, journey steps, or conditions on every current production trace. The platform renamed its observation spans (`preparation_iteration_N` → `Iteration: N`, `execute_tool_calls` → `Execute: Tools`, etc.) and the summary walker still matched only the legacy names, so it silently found nothing. `traces diff` builds on the same model, so it was comparing empty decision paths. On top of that, reviewing any window of activity meant fetching traces one at a time.

This PR repairs the summary and makes it usable at scale: one `traces list --summary` invocation now returns trustworthy, compact summaries for a whole filtered page of traces.

## Changes

- **Summary walker fixed** (`internal/summary`): matches both legacy engine span names and the current display names (payloads are identical across the two schemes, so no second parser). Verified against production traces from three projects — a KYC turn that previously summarized to nothing now shows all 8 iterations with journey steps, decision rationales, and tool calls with args/results.
- **WARNING visibility**: warning status lines (e.g. `no routine matched`) now appear in the summary's errors list instead of only ERROR — real issues at WARNING level no longer produce a false all-clear.
- **Correct iteration ordering**: iterations sort by span start time. Traces that run parallel sub-flows or restart the engine reuse iteration numbers; numeric ordering scrambled the story (one live trace holds 74 iteration spans across ~11 concurrent sub-runs).
- **`traces list --summary`**: composes with all existing list filters (`--level`, `--tags`, timestamps, cost/latency ranges, …); fetches per-trace detail + observations with a bounded worker pool (8) while preserving list order; a failed trace becomes a per-item error entry instead of aborting the batch; `--json`/`--yaml` emit structured summary models; mutually exclusive with `--columns`/`--fields`; pagination footer keeps truncation visible.
- **Docs**: regenerated `docs/iai_traces_list.md`.

Measured on production data: reviewing 10 error turns went from 10 sequential fetches to one command, ~2 s, ~30 KB of readable output — and the summary content is now complete enough to answer questions like "did that failed tool call eventually succeed before the agent replied?"

## Testing

- New live-shaped fixtures in `internal/summary`: display-name chat and autonomous flows, WARNING collection, repeated iteration numbers keeping chronological order; legacy-name cases retained as regression tests.
- New `cmd/traces_test.go`: batch summary over a stubbed API — order preservation, per-item errors, structured JSON shape, human rendering, and flag exclusions.
- `go test ./...` green except seven pre-existing timezone-dependent failures in `internal/output` (identical set fails on `main`).
- Smoke-tested against live Betsson and Alunafi projects.


internal/summary/trace.go
```diff
@@ -163,22 +184,46 @@ func indexTraceObservations(
...
 			n, _ := strconv.Atoi(sm[1])
-			iters = append(iters, iterNode{num: n, id: o.ID})
+			start, err := time.Parse(time.RFC3339Nano, o.StartTime)
+			iters = append(iters, iterNode{
+				num: n, id: o.ID, start: start, hasStart: err == nil,
+			})
 		}
 	}
-	sort.Slice(iters, func(i, j int) bool { return iters[i].num < iters[j].num })
+	sort.SliceStable(iters, func(i, j int) bool {
```
dgallego-iai, line 192: Repeated iteration numbers are preserved here, but `TraceDiff` later stores them by number (`internal/summary/diff.go:150`), replacing earlier entries. A sequence like `1,2,1,2` becomes `1,2`, so differences in earlier paths can disappear. Could we preserve every occurrence and add a test with repeated numbers?

cmd/traces.go
```diff
@@ -279,6 +296,39 @@ Uses the platform API with dual authentication (API key or session).`,
...
+	items := make([]summary.TraceSummaryItem, len(traces))
+	sem := make(chan struct{}, workers)
+	var wg sync.WaitGroup
+	for i, t := range traces {
+		wg.Add(1)
+		go func(i int, traceID string) {
+			defer wg.Done()
+			select {
+			case sem <- struct{}{}:
+			case <-ctx.Done():
```
dgallego-iai, line 318: When the request is cancelled, this records an item error, but `traceSummariesFor` can only return items, so the command later exits successfully. Could we return the cancellation error after the requests stop?

internal/summary/trace.go
```diff
@@ -163,22 +184,46 @@ func indexTraceObservations(
...
+	})
 	return children, iters, errs
 }
 
+func statusLine(o clients.ObservationInfo) string {
+	switch {
+	case strings.EqualFold(o.Level, "ERROR"):
+		return o.Name + ": " + valueOr(o.StatusMessage, "error")
+	case strings.EqualFold(o.Level, "WARNING"):
+		return "WARNING — " + o.Name + ": " + valueOr(o.StatusMessage, "warning")
```
dgallego-iai, line 209: Is it intentional that warnings are stored under `errors` and printed under `Errors:`? A warning-only trace may then look failed to users and scripts. If intentional, no change needed.

cmd/traces_test.go
```diff
@@ -0,0 +1,173 @@
...
+			tracesListCmd.Flags().Lookup(name).Changed = false
+		}
+		tracesListCmd.SetOut(nil)
+		tracesListCmd.SetErr(nil)
+	})
+	hostname, token, apiKey = server.URL, "test-token", ""
+	tracesListOrg, tracesListProject = "acme", "alunafi"
+}
+
+func TestTracesListSummaryJSON(t *testing.T) {
```
dgallego-iai, line 81: Could the JSON and human-output cases use one test table and compare the complete result with `cmp.Diff`? The current field and substring checks can miss missing, duplicated, or reordered output.

cmd/traces.go
```diff
@@ -415,11 +465,15 @@ func init() {
 	// Output flags
 	tracesListCmd.Flags().BoolVar(&tracesJSON, "json", false, "Output raw API response as JSON")
 	tracesListCmd.Flags().BoolVar(&tracesYAML, "yaml", false, "Output raw API response as YAML")
+	tracesListCmd.Flags().BoolVar(&tracesListSummary, "summary", false,
+		"Render a compact, LLM-readable summary of each listed turn; with --json/--yaml, emits structured summary models instead of the raw API response")
```
dgallego-iai: `LLM-readable` and `structured summary models` are internal terms in user help. Could we describe what users get instead, for example: “Show conditions, decisions, tools, and results for each listed turn”?

internal/summary/trace.go
```diff
@@ -163,22 +184,46 @@ func indexTraceObservations(
...
-				msg = "error"
-			}
-			errs = append(errs, o.Name+": "+msg)
+		if line := statusLine(o); line != "" {
+			errs = append(errs, line)
 		}
 		if sm := iterationNameRe.FindStringSubmatch(o.Name); sm != nil {
 			n, _ := strconv.Atoi(sm[1])
-			iters = append(iters, iterNode{num: n, id: o.ID})
+			start, err := time.Parse(time.RFC3339Nano, o.StartTime)
```
dgallego-iai, line 186: The comment above no longer matches this function: it now includes warnings and orders iterations by start time. Please update it or remove it.

internal/summary/trace.go
```diff
@@ -67,22 +74,34 @@ type TraceSummaryModel struct {
...
+		"Execute: Tools":     true,
+	}
+	kbRetrievalSpans = map[string]bool{
+		"retriever:knowledge_base": true,
+		"find_similar_documents":   true,
+		"Search: Documents":        true,
+	}
 )
 
+func isNextStepSpan(name string) bool {
```
dgallego-iai: This helper is one line and has one caller. Could we inline the condition at its use and remove the helper? Keeping the matching rule where it is used makes the code easier to read.

internal/summary/trace.go
oryall: Fixed in 933452d: `journeyLabelsByIter` now returns `map[int][][]string`, keeping every occurrence of a repeated iteration number, and `journeyDiff` emits one row per occurrence — a `1,2,1,2` sequence now yields all four entries. Added a `TestTraceDiff` case with repeated numbers.

cmd/traces.go
oryall: Fixed in 933452d: `traceSummariesFor` now returns an error — after the workers stop it checks `ctx.Err()` and returns it, so a cancelled run fails the command instead of printing partial items and exiting 0. Covered by `TestTraceSummariesForReturnsCancellation`.

internal/summary/trace.go
oryall: Intentional — surfacing WARNING findings (e.g. `no routine matched`) is the point of the change, since they previously produced a false all-clear. Each warning line carries a `WARNING — ` prefix so it stays distinguishable from errors, and the trace-level `level` field still reports the aggregated severity, so a warning-only trace is not reported as failed.

cmd/traces_test.go
oryall: Addressed in 933452d: both cases now compare the complete result with `cmp.Diff` — the JSON test diffs the full decoded `[]TraceSummaryItem` slice and the human test diffs the entire rendered output against a golden string, so missing, duplicated, or reordered output now fails. Kept them as two functions sharing one stub server rather than one table, since they assert different types (struct slice vs rendered text) and need different flag setups.

cmd/traces.go
oryall: Done in 933452d — the flag help now reads: "Show conditions, decisions, tools, and results for each listed turn". `docs/iai_traces_list.md` regenerated to match.

internal/summary/trace.go
oryall: Removed in 7552995, together with the other docstrings this PR had added.

internal/summary/trace.go
oryall: Done in 933452d — `isNextStepSpan` is removed and the condition is inlined at its single use in `summarizeIteration`.

dgallego-iai (approved):

PR #179, opened 2026-08-05
Add stacks get, diff, and list commands (DEV-944)

dgallego-iai:
## What

Three new `iai stacks` subcommands for the Terraform-plan workflow: export, diff, and list.

### `iai stacks get`
Export live stack configuration as YAML (services + agents + databases).

```bash
iai stacks get --stack-id my-stack
iai stacks get --stack-id my-stack -f live.yaml
```

### `iai stacks diff`
Compare a local config file against the live stack. Shows creates, updates (with field-level diffs), and deletes. Supports `--json` for CI.

```bash
iai stacks diff --file stack.yaml --stack-id my-stack
iai stacks diff --file stack.yaml --stack-id my-stack --json
```

**Human output** uses a scannable plan format:

```
Stack: my-stack

  + service worker (create)
  ~ service api (update)
    replicas: 1 → 2
    resources.memory: 256M → 512M
    + env[0].name: LOG_LEVEL
  - service old-svc (delete)
```

**JSON output** includes field-level change details for CI consumption.

### `iai stacks list`
Discover stacks and their resource counts in a project.

### New flags
- `stacks get`: `--stack-id`, `-f/--file`, `-o/--organization`, `-p/--project`
- `stacks diff`: `--file`, `--stack-id`, `-o/--organization`, `-p/--project`, `--json`
- `stacks list`: `-o/--organization`, `-p/--project`, `--json`

## Changes
| File | Change |
|------|--------|
| `cmd/stack.go` | +241: three new commands with flag wiring |
| `internal/files/stack_config.go` | +72: conversion functions (API → config) + json tags |
| `internal/files/stack_live.go` | new: FetchLiveStack, MarshalStackConfig |
| `internal/files/stack_list.go` | new: ListStacks (discover stacks from live resources) |
| `internal/files/stack_diff.go` | new: DiffStackConfigs, field-level diff, plan output |
| `internal/files/stack_config_test.go` | +96: table-driven tests for conversion functions |
| `internal/files/stack_diff_test.go` | new: table-driven tests for diff, flatten, plan output |

Closes DEV-944


dgallego-iai:
All review items addressed:

**Must fix:**
- ✅ Dead code removed (`serviceConfigEqual`, `agentConfigEqual`, `databaseConfigEqual` + yaml import)
- ✅ `ListStacks` output now sorted by StackID

**Should fix:**
- ✅ Double `MarshalStackConfig` call removed — file-write branch reuses `yamlData`
- ✅ Misleading "(or against another file)" removed from diff description

**Informational:**
- ✅ N+1 comment removed (per repo style — addressed here instead)
- ✅ Comment added explaining why `diffFields` is re-run in `printSection`

**Additional:**
- Added `--json`/`--yaml` flags to `stacks get` for consistency
- Added `json` tags to `StackConfig` top-level fields
- Pre-commit hooks pass (gofumpt, golines, doc generation)

dgallego-iai:
Per-discussion status:

- **#r3719418575** (dead code): ✅ removed in 3be8633
- **#r3719418689** (non-deterministic sort): ✅ sorted by StackID in 3be8633  
- **#r3719418080** (double MarshalStackConfig): ✅ reuses yamlData in 3be8633
- **#r3719419528** (misleading "or against another file"): ✅ removed in 3be8633
- **#r3719419409** (N+1 describes): ✅ addressed via PR comment (no code comment per repo style)
- **#r3719420270** (double diffFields): ✅ explainer comment added in 3be8633

dgallego-iai:
@claude please re-review — all previous items addressed, e2e test battery passes (20/20).

cmd/stack.go
```diff
@@ -264,6 +282,203 @@ The organization and project are read from the config file, flags, or resolved v
...
+		if filePath == "" {
+			return fmt.Errorf("config file is required; please provide --file or --cfg-file")
+		}
+
+		localCfg, err := files.LoadStackConfig(filePath)
+		if err != nil {
+			return fmt.Errorf("failed to load local config: %w", err)
+		}
+
+		pCtx, _, deployClient, err := resolveProject(
```
oryall, line 392: `diff` resolves org/project only from flags/session — it never looks at the `organization:`/`project:` fields in the `--file` config, unlike `sync` which does. So if my session has a different org selected, `sync --file stack.yaml` targets the file's org but `diff --file stack.yaml` silently diffs against the wrong project and reports everything as created/deleted. Can we fall back to the file's org/project when flags aren't set, like sync?

internal/files/stack_config.go
```diff
@@ -111,6 +112,51 @@ func (d DatabaseConfig) ToCreateRequest(stackId string) clients.CreateDatabaseBo
 	}
 }
 
+func ServiceConfigFromDescribe(svc *clients.DescribeServiceResponse) ServiceConfig {
```
oryall, line 115: Any service config with `version:` set will show `+ version: v1` in the diff forever: the describe API has no version field, so the live side is always empty. Since `ToCreateRequest` doesn't send version either, sync never resolves it. Diff can never reach "No differences found" for those configs, which breaks the CI use case.

cmd/stack.go
dgallego-iai: Fixed in 667982f — diff now falls back to the config file's organization/project when -o/-p aren't set, matching sync behavior. The flag still wins if explicitly passed.

internal/files/stack_config.go
dgallego-iai: Fixed in 667982f — version is now filtered from field-level diffs since the describe API never returns it. Added TestDiffFieldsVersionFiltered to prevent regression. Configs with version: set can now reach 'No differences found.'

oryall (approved):

PR #184, opened 2026-08-18
feat(DEV-989): add queues delete and unblock item deletion for session callers

gabrielgustavoandrade:
## What changed

- New `iai queues delete <id>` (alias `rm`), following the `queue-items delete` shape.
- `DeleteQueueItem` no longer requires API-key auth on the client side.
- `ExtractServerMessage` understands the platform's `data.message` envelope.

## Why

DEV-956 asked for item removal and whole-queue deletion to work "both in-platform and via API… so Copilot can do them too". Neither was reachable from Copilot:

- There was no `queues delete` command at all. My note on #183 said it was omitted because deletion 403s for API-key auth — that was reasoning from the wrong auth mode. `backend/api/platform/v1/annotation_queues.py:245` rejects only `AuthMode.SDK`; Cookie and Bearer are served.
- `queue-items delete` called `requireAPIKeyMode()`, so Bearer callers were refused by the CLI even though the API handles both modes (`annotation_queues.py:514` branches SDK → public path, else → tRPC mutation).

Copilot authenticates with a Bearer token from the sandbox's `device_login.sh` (its secret holds only `SANDBOX_AUTH_TOKEN`; there is no `INTERACTIVE_API_KEY`), so both paths were dead for it. Before this change, in session mode:

```
$ iai queue-items delete <item> --queue-id <queue>
Error: this command requires API key authentication (--api-key or INTERACTIVE_API_KEY)
```

The `data.message` fix is not cosmetic drive-by: delete responses wrap the message under `data.message`, which matched none of the error shapes and fell through to the raw-body fallback, so `PrintDeleteSuccess` printed the entire JSON envelope. The new command would have inherited that.

## Verified against dev in session/Bearer mode

Created a throwaway queue in the dev UI, then deleted it with this build:

```
$ iai queues delete cmsyqrxth04f6zd07dnjt4j98 -o platform-development -p cli-tests
Annotation queue deleted
exit=0

$ iai queues list -o platform-development -p cli-tests --columns id,name
ID                          NAME
cmn7nh1tc00ffwg07k9veidlp   review-queue
cmn7ng92s00f7wg07l3rf93f4   test-queue-2

$ iai queues delete cmsyqrxth04f6zd07dnjt4j98 ...        # deleting again
Error: Annotation queue 'cmsyqrxth04f6zd07dnjt4j98' not found
```

## Tests

`TestAPIClientDeleteAnnotationQueue` asserts method, path and the extracted message. `TestAPIClientDeleteQueueItemWithoutAPIKey` pins the ungating — it fails if the API-key gate returns, since the test client is in Bearer mode.

## Still gated on API keys, deliberately

`queues create`, `queues assign` / `unassign`, `queue-items create` and `queue-items update` keep `requireAPIKeyMode()`. Item update is genuinely mode-dependent — session callers may only complete an item (`annotation_queues.py:443`) — and the others are outside what DEV-956 asked for. Worth a follow-up look if Copilot needs to create queues.

### DIFF
--- a/internal/clients/api_client_eval_test.go
+++ b/internal/clients/api_client_eval_test.go
@@ -168,69 +168,63 @@ func TestAPIClientListQueueItems(t *testing.T) {
 	}
 }
 
-func TestAPIClientDeleteAnnotationQueue(t *testing.T) {
-	wantPath := "/api/platform/v1/organizations/org-1/projects/proj-1/annotation-queues/q-1"
-	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if r.Method != http.MethodDelete {
-			t.Errorf("method = %s, want DELETE", r.Method)
-		}
-		if r.URL.Path != wantPath {
-			t.Errorf("path = %s, want %s", r.URL.Path, wantPath)
-		}
-		_, _ = io.WriteString(
-			w,
-			`{"success":true,"data":{"message":"Annotation queue deleted"}}`,
-		)
-	}))
-	defer server.Close()
+func TestAPIClientDeleteQueueResources(t *testing.T) {
+	const base = "/api/platform/v1/organizations/org-1/projects/proj-1/annotation-queues"
 
-	client := newEvalTestClient(t, server.URL)
-	message, err := client.DeleteAnnotationQueue(
-		context.Background(),
-		"org-1",
-		"proj-1",
-		"q-1",
-	)
-	if err != nil {
-		t.Fatalf("DeleteAnnotationQueue() error = %v", err)
+	tests := []struct {
+		name        string
+		wantPath    string
+		body        string
+		wantMessage string
+		call        func(*APIClient) (string, error)
+	}{
+		{
+			name:        "delete annotation queue",
+			wantPath:    base + "/q-1",
+			body:        `{"success":true,"data":{"message":"Annotation queue deleted"}}`,
+			wantMessage: "Annotation queue deleted",
+			call: func(c *APIClient) (string, error) {
+				return c.DeleteAnnotationQueue(context.Background(), "org-1", "proj-1", "q-1")
+			},
+		},
+		{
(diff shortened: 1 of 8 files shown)


gabrielgustavoandrade:
@claude both applied in `d1dc7d8`:

- `TestAPIClientDeleteQueueItemWithoutAPIKey` now asserts `r.Method == http.MethodDelete`, matching the adjacent test.
- The `ExtractServerMessage` doc comment lists five formats, with the platform success envelope (`{"data": {"message": "..."}}`) as the new #4 ahead of the plain-text fallback.

@dgallego-iai this one is on you when you have a moment — it closes the deletion half of DEV-956 that #183 left out. Verified end to end on dev in session/Bearer mode: created a throwaway queue in the UI, deleted it with `iai queues delete`, confirmed it was gone and that a repeat delete 404s cleanly.

internal/clients/api_client_eval_test.go
```diff
@@ -167,3 +167,70 @@ func TestAPIClientListQueueItems(t *testing.T) {
 		})
 	}
 }
+
+func TestAPIClientDeleteAnnotationQueue(t *testing.T) {
```
dgallego-iai: Can we combine the two standalone tests into a table-driven delete test?

gabrielgustavoandrade: Done in `ff099b3`. The two standalone tests are now one table-driven `TestAPIClientDeleteQueueResources` with a `call func(*APIClient) (string, error)` field per case, so the shared server setup, the DELETE-method assertion and the path/message checks live in one place:

- `delete annotation queue` → `.../annotation-queues/q-1`
- `delete queue item without an API key` → `.../annotation-queues/q-1/items/i-1`

The bearer-mode assertion moved into the shared body and now covers both cases, which is a small improvement — it pins the ungating for the item delete and also guards the new queue delete against someone adding an API-key gate to it later.

dgallego-iai (approved):

